PR #80, opened 2023-06-23
Actor critic tests

dlangbe:
Adds the following tensor contraction tests using the actor-critic contraction selection algorithm:
- bilinear_contraction_ac_f32
- bilinear_contraction_ac_f64
- scale_contraction_ac_f32
- scale_contraction_ac_f64

Adds timing for contraction selection and modifies kernel performance logging.
- hiptensorContraction will record and log timing and throughput results when logging mode is set to `HIPTENSOR_LOG_LEVEL_PERF_TRACE`


library/src/contraction/contraction_selection.cpp
```diff
@@ -2602,7 +2595,18 @@ namespace hiptensor
 
             if(auto candidate = candidates.find(unique_id); candidate != candidates.end())
             {
-                *winner = candidate->second;
+                auto* solution = candidate->second;
+
+                PerfMetrics metrics = {
```
cgmillette: Can probably remove the metrics as they are unused. Then can condense the assignment of winner to candidate->second

cgmillette (approved):

PR #93, opened 2023-07-26
Initial yaml-config

cgmillette:
- Use existing llvm libs with clang deployment
- Setup initial schema for Contraction Test
- Add utils for read / write to file
- Add object for YAML wrappers.
- Fixup some project hierarchy


dlangbe (approved):
LGTM!

PR #102, opened 2023-08-30
Add Doxygen warnings as errors

CongMa13:
- There are bugs in Doxygen 1.8.13 which is used by readthedocs
  - Skip code to avoid Doxygen errors


cgmillette (approved):
LGTM!

PR #107, opened 2023-09-06
Fix compiler errors in debug build

cgmillette:
hiptensorPrintArrayElements function interface has changed


CongMa13 (approved):
lgtm. 

Just curious, wasn't there build error for PR that changed hiptensorPrintArrayElements?

cgmillette:
There might have been, however the CI only runs a release build atm. Maybe we can look at adding pipes for the debug builds as well.
The instances causing the build error were hidden behind !NDEBUG

PR #108, opened 2023-09-06
Update tests

cgmillette:
- Reinstate some meaningful test sizes that are compatible with hipTensor.
- Separate contraction_test into hpp/cpp
- Buffer hipTensor API logging,
- Fix llvm command line option for showing hipTensor version
- Add llvm command line option for omitting outputs
- Control output flow with omit options
- Don't output tensors contents by default.
- Skip tests if device or run checks fail
- Add supported device check before startup


mkarunan (dismissed):
Looks good

dlangbe (dismissed):

PR #112, opened 2023-09-08
Add codeowners for documentation

samjwu:
Adding code owners for documentation. Plan is to add mandatory reviews from this group of code owners to merge any documentation changes. The header directory is related to documentation and thus changes to that are also flagged for documentation review.


cgmillette (changes requested):
Can you append this to the existing codeowners?
Otherwise this wipes out the existing settings too.

samjwu:
> Can you append this to the existing codeowners? Otherwise this wipes out the existing settings too.

Oops. Thanks for catching that.

cgmillette (approved):

PR #128, opened 2023-10-04
yaml_test saves test data into a temp file

CongMa13:
yaml_test will fail when the work folder is readonly, if it creates a file in the work folder

Thus, yaml_test saves test data into a temp file


cgmillette (approved):
LGTM

dlangbe (approved):

PR #145, opened 2023-10-31
Add CPU implementation of elementwise permutation

CongMa13:
- Add CPU implementation of elementwise permutation -  permutationByCpu
- Add sample of elementwise permutation

GPU implementation of elementwise permutation is not done yet.

### DIFF
--- a/library/src/permutation/permutation_cpu_reference.hpp
+++ b/library/src/permutation/permutation_cpu_reference.hpp
@@ -23,30 +23,28 @@
  * THE SOFTWARE.
  *
  *******************************************************************************/
+
+#ifndef HIPTENSOR_PERMUTATION_CPU_REFERENCE_HPP
+#define HIPTENSOR_PERMUTATION_CPU_REFERENCE_HPP
+
+#include <hip/library_types.h>
+
 #include <hiptensor/hiptensor.hpp>
-
-#include "permutation_cpu_reference.hpp"
-
-hiptensorStatus_t hiptensorPermutation(const hiptensorHandle_t*           handle,
-                                       const void*                        alpha,
-                                       const void*                        A,
+namespace hiptensor
+{
+    namespace detail
+    {
+        template <typename DataType>
+        hiptensorStatus_t permuteByCpu(const void*                        alpha,
+                                       const DataType*                    A,
                                        const hiptensorTensorDescriptor_t* descA,
                                        const int32_t                      modeA[],
-                                       void*                              B,
+                                       DataType*                          B,
                                        const hiptensorTensorDescriptor_t* descB,
                                        const int32_t                      modeB[],
-                                       const hipDataType                  typeScalar,
-                                       const hipStream_t                  stream)
-{
-    assert(descA->mType == HIP_R_16F || descA->mType == HIP_R_32F);
-    assert(descA->mType == descB->mType);
-    if(descA->mType == HIP_R_16F)
-    {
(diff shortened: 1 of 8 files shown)


library/src/permutation/permutation_cpu_reference.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef HIPTENSOR_PERMUTATION_CPU_REFERENCE_HPP
+#define HIPTENSOR_PERMUTATION_CPU_REFERENCE_HPP
+
+#include <hip/library_types.h>
+
+#include <hiptensor/hiptensor.hpp>
+namespace hiptensor
+{
+    namespace details
```
cgmillette: This namespace for consistency perhaps change to 'detail' instead of 'details'

library/src/permutation/hiptensor_permutation.cpp
```diff
@@ -36,5 +38,5 @@ hiptensorStatus_t hiptensorPermutation(const hiptensorHandle_t*           handle
                                        const hipDataType                  typeScalar,
                                        const hipStream_t                  stream)
 {
-    return HIPTENSOR_STATUS_SUCCESS;
+    return hiptensor::details::permuteByCpu(alpha, A, descA, modeA, B, descB, modeB, typeScalar);
```
cgmillette: This function will deduce DataType template argument as 'void'. Is this intended?

library/src/permutation/permutation_cpu_reference_impl.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "permutation_cpu_reference.hpp"
+#include "types.hpp"
+#include "util.hpp"
+
+namespace hiptensor
+{
+    namespace details
+    {
+        template <typename DataType>
```
cgmillette, line 40: Template type DataType doesn't appear to be used. Can we remove the template param and use void data pass-through from the API? E.g.
```
hiptensorStatus_t permuteByCpu(const void*                        alpha,
                                       const void*                    A,
                                       const hiptensorTensorDescriptor_t* descA,
                                       const int32_t                      modeA[],
                                       void*                          B,
                                       const hiptensorTensorDescriptor_t* descB,
                                       const int32_t                      modeB[],
                                       const hipDataType                  typeScalar)
```

samples/02_permutation/permutation.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include <hiptensor/hiptensor.hpp>
+#include <hiptensor/internal/hiptensor_utility.hpp>
+
+#include "../common.hpp"
```
cgmillette: Might be good to add the samples base dir to the include directories to avoid the relative paths

samples/common.hpp
```diff
@@ -35,10 +35,10 @@
...
     }
 
-#define HIPTENSOR_FREE_HOST(ptr) \
-    if(ptr != nullptr)           \
-    {                            \
-        free(ptr);               \
+#define HIPTENSOR_FREE_HOST(ptr)           \
+    if(ptr != nullptr)                     \
+    {                                      \
+        CHECK_HIP_ERROR(hipHostFree(ptr)); \
```
cgmillette, line 41: Nice catch

library/src/permutation/hiptensor_permutation.cpp
cgmillette: See comment below on the permuteByCpu function

cgmillette (approved):
Lgtm

PR #148, opened 2023-11-02
Replace sprintf with snprintf

CongMa13:
snprintf is safe function since it will not write data beyond the limited size.


cgmillette (approved):
Looks good thanks for the catch

PR #157, opened 2023-11-21
Enable  Contraction for complex float and double types

mkarunan:
Enable complex float and double support in library.
Modify CPU Contraction reference class to handle complex types.
Add CPU scale contraction samples for complex types.


library/include/hiptensor/internal/hiptensor_utility.hpp
```diff
@@ -60,6 +61,16 @@
...
+inline std::ostream& operator<<(std::ostream& os, const hipFloatComplex& fc)
+{
+    return os << hipCrealf(fc) << " " << hipCimagf(fc);
+}
+
+inline std::ostream& operator<<(std::ostream& os, const hipDoubleComplex& dc)
+{
+    return os << hipCreal(dc) << " " << hipCimag(dc);
+}
+
```
CongMa13: Could we print complex in a format like "3 + 2i"?

library/src/hiptensor.cpp
```diff
@@ -151,8 +151,7 @@ hiptensorStatus_t hiptensorInitTensorDescriptor(const hiptensorHandle_t*     han
         return HIPTENSOR_STATUS_NOT_INITIALIZED;
     }
 
-    if((lens == nullptr)
-       || ((dataType != HIP_R_16F) && (dataType != HIP_R_32F) && (dataType != HIP_R_64F))
+    if((lens == nullptr) || ((dataType != HIP_R_32F) && (dataType != HIP_R_64F) && (dataType != HIP_C_32F) && (dataType != HIP_C_64F))
```
CongMa13: It's nice to support HIP_C_32F and HIP_C64F. And could we keep the support to HIP_R_16F as well since permute has been supporting f16. Permute_test cannot pass without it.

samples/01_contraction/simple_scale_contraction_doubleComplex_cpu.cpp
```diff
@@ -0,0 +1,232 @@
...
+                                     ((double(std::rand())) / double(RAND_MAX) - 0.5) * 100);
+    }
+
+    for(int64_t i = 0; i < elementsD; i++)
+    {
+        D[i] = make_hipDoubleComplex((double)std::numeric_limits<double>::signaling_NaN(),
+                                     (double)std::numeric_limits<double>::signaling_NaN());
+    }
+
+    CHECK_HIPTENSOR_ERROR(hiptensorContractionReference((void*)&alpha,
```
CongMa13, line 159: Just want to confirm that this will be replaced with CK implementation later?

CongMa13 (changes requested):

samples/01_contraction/simple_scale_contraction_doubleComplex_cpu.cpp
mkarunan: yes right, this sample will be a device version like other samples for f32 , f64. Right now, its only a cpu version for complex float and double.

test/01_contraction/contraction_test.cpp
```diff
@@ -328,7 +354,9 @@ namespace hiptensor
             {
                 auto resource = getResource();
 
-                int size = ((DDataType == HIP_R_32F) ? sizeof(float) : sizeof(double));
+                int size = ((DDataType == HIP_R_32F) ? sizeof(float) : (DDataType == HIP_C_32F)
```
cgmillette, line 357: ~~Might be a good idea to implement a utility function to determine the element size from the hipDataType~~

Please use hipDataTypeSize utility here.

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -95,11 +95,11 @@ namespace hiptensor
 
             if(alpha != nullptr)
             {
-                alphaF = hiptensor::readVal<float>(alpha, HipDataType_v<typename Traits::EDataT>);
+                alphaF = hiptensor::readVal<float>(alpha, convertToComputeType(HipDataType_v<typename Traits::EDataT>));
```
cgmillette, line 132: Compute type should be available in MetaTraits<DeviceOp>::ComputeT

test/01_contraction/contraction_test.cpp
```diff
@@ -367,6 +395,48 @@ namespace hiptensor
                     hiptensorPrintArrayElements<float>(stream, (float*)D.get(), elementsCD);
                     stream << std::endl;
                 }
+                else if(DDataType == HIP_C_32F)
+                {
+                    stream << "Tensor A elements:\n";
```
cgmillette, line 400: The type of D shouldn't determine the type of the rest of ABC.
What we need to do here is implement a wrapper:
```
hipTensorPrintArrayElements(stream, void* data, elementCount, hipDataType dataType)
{
    if (dataType == HIP_R_32F)
    {
        hipTensorPrintArrayElements<float>(...);
    } else if...
}
```

That way the dispatching is wrapped up. The new code should look something like
```
hipTensorPrintArrayElements(stream, resource->hostA().get(), elementsA, ADataType);
hipTensorPrintArrayElements(stream, resource->hostB().get(), elementsA, BDataType);
...
```

test/01_contraction/contraction_test.cpp
```diff
@@ -456,16 +526,19 @@ namespace hiptensor
                                                 size_t{1},
                                                 std::multiplies<size_t>());
 
-            int  sizeD = elementsCD * ((DDataType == HIP_R_32F) ? sizeof(float) : sizeof(double));
+            int sizeD = elementsCD * ((DDataType == HIP_R_32F) ? sizeof(float) : (DDataType == HIP_C_32F)
```
cgmillette, line 529: use the hipDataSize utility

test/device/common.hpp
```diff
@@ -72,8 +72,21 @@ __global__ void fillKernel(DataType* data, uint32_t elementSize, uint32_t seed)
 
     if(index < elementSize)
     {
-        auto value  = (DataType(index / DataType(RAND_MAX) - 0.5) * 100) / elementSize;
-        data[index] = static_cast<DataType>(value);
+        if constexpr(std::is_same_v<DataType, hipFloatComplex>)
+        {
+            auto value  = (float(index / float(RAND_MAX) - 0.5) * 100) / elementSize;
+            data[index] = DataType(value, value);
```
cgmillette, line 78: Since this check is explicitly for hipFloatComplex, then we should use the factory functions:
make_hipFloatComplex
make_hipDoubleComplex

test/01_contraction/contraction_test.cpp
```diff
@@ -456,16 +526,19 @@ namespace hiptensor
...
+                                                               ? sizeof(hipFloatComplex) : (DDataType == HIP_C_64F)
+                                                               ? sizeof(hipDoubleComplex) : sizeof(double));
+
             auto reference = resource->allocDevice(sizeD);
             resource->copyData(reference, resource->hostD(), sizeD);
 
-            if(DDataType == HIP_R_32F)
+            if(DDataType == HIP_R_32F || DDataType == HIP_C_32F)
             {
                 std::tie(mValidationResult, mMaxRelativeError) = compareEqualLaunchKernel<float>(
```
cgmillette, line 538: Same as with hipTensorPrintArrayElements, we might want another wrapper so that we can hide the dispatching here in the wrapper.

test/01_contraction/contraction_test.cpp
```diff
@@ -241,6 +241,19 @@ namespace hiptensor
                                            elementsCD,
                                            std::numeric_limits<float>::signaling_NaN());
             }
+            else if(ADataType == HIP_C_32F && BDataType == HIP_C_32F && DDataType == HIP_C_32F)
```
cgmillette, line 244: See comments for hipTensorPrintArrayElements

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -46,20 +46,68 @@ namespace std
 
 namespace hiptensor
 {
-    template <typename DeviceOp, typename Enabler = void>
+    template <typename DeviceOp, typename DataType, typename Enabler = void>
```
cgmillette, line 49: You don't need to add DataType as a template parameter here. 
You can get the type information you need below from the typename MetaTraits<DeviceOp>::ADataT

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -232,10 +389,723 @@ namespace hiptensor
...
+            D_d_imag.reset(nullptr);
+            D_d_imag = std::move(allocDevice<elementType>(elementsD));
+            E_d_real.reset(nullptr);
+            E_d_real = std::move(allocDevice<elementType>(elementsE));
+            E_d_imag.reset(nullptr);
+            E_d_imag = std::move(allocDevice<elementType>(elementsE));
+
+            auto blockDim = dim3(1024, 1, 1);
+            auto gridDim  = dim3(ceilDiv(elementsA, blockDim.x), 1, 1);
+            hipLaunchKernelGGL(
```
cgmillette: We usually use hipLaunchKernelGGL if we need timing params through events. These are unused here so we can get away with the triple chevron syntax instead <<<>>>

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -232,10 +389,723 @@ namespace hiptensor
...
+                            (unpack<DataType, elementType>), gridDim, blockDim, 0, 0, (const DataType*)D,
+                            (elementType*)D_d_real.get(), (elementType*)D_d_imag.get(), elementsD);
+
+            gridDim  = dim3(ceilDiv(elementsE, blockDim.x), 1, 1);
+            hipLaunchKernelGGL(
+                            (unpack<DataType, elementType>), gridDim, blockDim, 0, 0, (const DataType*)E,
+                            (elementType*)E_d_real.get(), (elementType*)E_d_imag.get(), elementsE);
+
+            // Initialize the argument pointer
+            mArgPtr[0] = std::move(deviceOp->MakeArgumentPointer(
```
cgmillette: You might want to create a lambda helper function to create each argument pointer so reduce some of the repetition here. The input args to the lambda would be the items that change (device ptrs, etc). The rest would be captures (such as lengths / strides)

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -232,10 +389,723 @@ namespace hiptensor
...
+                toCKVec(e_ms_ns_strides),
+                typename Traits::AOp{},
+                typename Traits::BOp{},
+                typename Traits::CDEOp{alphaF, betaF}));
+
+            // Attach the workspace pointer
+            deviceOp->SetWorkSpacePointer(mArgPtr[0].get(), workspacePtr);
+
+            // Initialize the invoker
+            mInvokerPtr[0] = std::move(deviceOp->MakeInvokerPointer());
```
cgmillette: Not sure that there is a need for 4 separate invokers because they are all the same kernel. We can try to feed each of the separate args kernels to just a single invoker ptr.

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -232,10 +389,723 @@ namespace hiptensor
...
+        : public ContractionSolution
+    {
+    protected :
+        std::vector<std::unique_ptr<ck::tensor_operation::device::BaseArgument>> mArgPtr;
+        std::vector<std::unique_ptr<ck::tensor_operation::device::BaseInvoker>>  mInvokerPtr;
+
+        using elementType = typename std::conditional<std::is_same<DataType, hipFloatComplex>::value,
+                                                      float, double>::type;
+
+        std::unique_ptr<elementType, DeviceDeleter> A_d_real;
```
cgmillette, line 490: Please prefix member variables with m

library/src/contraction/contraction_solution_impl.hpp
```diff
@@ -232,10 +389,723 @@ namespace hiptensor
...
+        std::unique_ptr<elementType, DeviceDeleter> A_d_imag;
+        std::unique_ptr<elementType, DeviceDeleter> B_d_real;
```
cgmillette, line 492: You may want to define an alias for the device allocations
```
template<typename DataT>
using DevicePtrT = std::unique_ptr<DataT, DeviceDeleter>;

DevicePtrT mAReal;
DevicePtrT mAImag;
...
```

This will help to cut down on repetition

cgmillette (changes requested):
In general, I think the approach is sound, however there are some areas we can de-risk and improve code quality. Please see other comments!

PR #160, opened 2023-11-29
Fix docs build error and warnings

neon60:
@parbenc please continue, fix all warnings.

### DIFF
--- a/library/include/hiptensor/internal/type_traits.hpp
+++ b/library/include/hiptensor/internal/type_traits.hpp
@@ -271,66 +271,6 @@ namespace std
         hiptensor::detail::Fp16Bits eps(static_cast<uint16_t>(0x7FC0));
         return eps.b16;
     }
-
-    ///////////////////////////////////////////////////////////
-    ///////////  std::numeric_limits<xfloat32_t>  //////////////
-    ///////////////////////////////////////////////////////////
-
-    template <>
-    HIPTENSOR_HOST_DEVICE constexpr hiptensor::xfloat32_t
-        numeric_limits<hiptensor::xfloat32_t>::epsilon() noexcept
-    {
-        hiptensor::detail::Fp32Bits eps(static_cast<float>(FLT_EPSILON));
-        return eps.xf32;
-    }
-
-    template <>
-    HIPTENSOR_HOST_DEVICE constexpr hiptensor::xfloat32_t
-        numeric_limits<hiptensor::xfloat32_t>::infinity() noexcept
-    {
-        hiptensor::detail::Fp32Bits eps(static_cast<float>(HUGE_VALF));
-        return eps.xf32;
-    }
-
-    template <>
-    HIPTENSOR_HOST_DEVICE constexpr hiptensor::xfloat32_t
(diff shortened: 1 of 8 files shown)


neon60:
@samjwu There are some overlap between your spelling branch and our branch. Can you check our changes and suggest if we should take over your commit(s)? We can also add the changes to your branch, if you want.

samjwu:
> @samjwu There are some overlap between your spelling branch and our branch. Can you check our changes and suggest if we should take over your commit(s)? We can also add the changes to your branch, if you want.

I think it might be clearest to have both PRs to focus on separate concerns. It looks like this one fixes some things with doxygen while mine only does spelling.

samjwu (dismissed):

neon60:
> > @samjwu There are some overlap between your spelling branch and our branch. Can you check our changes and suggest if we should take over your commit(s)? We can also add the changes to your branch, if you want.
> 
> I think it might be clearest to have both PRs to focus on separate concerns. It looks like this one fixes some things with doxygen while mine only does spelling.

You are right. We will have separate PRs and your PR already merged in.

parbenc:
Separated spellcheck stuff into another branch.

neon60:
> Separated spellcheck stuff into another branch.

@parbenc Can you open that PR?

neon60:
@parbenc Can we remove the WIP from the title?

library/include/hiptensor/internal/type_traits.hpp
```diff
@@ -270,6 +271,68 @@ namespace std
         hiptensor::detail::Fp16Bits eps(static_cast<uint16_t>(0x7FC0));
         return eps.b16;
     }
+
+    ///////////////////////////////////////////////////////////
```
neon60: Seems to be new functions added? Do we need these?

parbenc: you're right, this has sneaked back with the rebase. i have removed it

parbenc:
> @parbenc Can we remove the WIP from the title?

Unfortunately i can't edit this PR.

neon60:
Looks good to me.

parbenc:
@cgmillette could you please look at this?

library/include/hiptensor/internal/type_traits.hpp
```diff
@@ -270,6 +271,8 @@ namespace std
         hiptensor::detail::Fp16Bits eps(static_cast<uint16_t>(0x7FC0));
         return eps.b16;
     }
+#endif // DOXYGEN_SHOULD_SKIP_THIS
+    // @endcond
```
cgmillette: Need the endcond?

cgmillette (commented):
LGTM - just not sure about the dangling endcond

library/include/hiptensor/internal/type_traits.hpp
parbenc: probably not. it snuck back in during a rebase

parbenc: removed it. could you please approve it again

cgmillette (approved):

saadrahim (approved):

PR #165, opened 2023-12-07
Update llvm raw_fd_ostream usage to new interface

cgmillette:
The constructor for raw_fd_ostream has changed with the new llvm version.

Adjusted to use default of  llvm::sys::fs::F_None flag.


CongMa13 (approved):
lgtm

CongMa13 (changes requested):
Seems like there is missing header file.

cgmillette:
> Seems like there is missing header file.

Yea, because it's a breaking change with new llvm version. CI not updated to it yet

PR #172, opened 2023-12-19
New single kernel selection. To be improved.

afanfa:
All new cases now have only one option as "best" kernel. Needs to be improved. The selected kernels are compatible with all sizes tested and behave ok (but not optimally) with most of the sizes.


CongMa13:
> All new cases now have only one option as "best" kernel. Needs to be improved. The selected kernels are compatible with all sizes tested and behave ok (but not optimally) with most of the sizes.

Great, thanks for the update. Could we enable `HIPTENSOR_ALGO_ACTOR_CRITIC` in test/01_contraction/configs/*_test_params.yaml since the selected instance can handle all sized tests now?

cgmillette:
> > All new cases now have only one option as "best" kernel. Needs to be improved. The selected kernels are compatible with all sizes tested and behave ok (but not optimally) with most of the sizes.
> 
> Great, thanks for the update. Could we enable `HIPTENSOR_ALGO_ACTOR_CRITIC` in test/01_contraction/configs/*_test_params.yaml since the selected instance can handle all sized tests now?

Yeah, I think we can push that change with this PR

CongMa13:
This commit has been cherry-picked to another PR #174. So we will close this PR.

PR #168, opened 2023-12-11
Adding issue template

abhimeda:
Hello, this PR is introducing a new issue template for ROCm public repositories to make it easier for users to submit the details for issues and bugs. Problems pertaining to documentation or non-bug related, users have the ability to open a blank issue template to submit their report. 

Thank you,
Community Support Team (ML SW SDK)

PR #177, opened 2024-01-05
Revert "Adding issue template"

cgmillette:
Reverts ROCm/hipTensor#168
@CongMa13:  Can we please check if this made it to staging branch?  @abhimeda has informed that we should revert.


CongMa13:
@cgmillette 

#168 is not in release-staging/rocm-rel-6.1 so that we only need to revert it in develop br.

PR #179, opened 2024-01-24
Add flags needed for clang compiler

mkarunan:
To transition from hipcc to clang, add additional flags to match the compiler flags generated from hipcc.


cgmillette (changes requested):
This makes the changes global to the current project.
Can we instead make these changes to the test and samples CMakeLists for the target creation:
target_compile_options($target PRIVATE --driver-mode=g++ )

CongMa13:
> This makes the changes global to the current project. Can we instead make these changes to the test and samples CMakeLists for the target creation: target_compile_options($target PRIVATE --driver-mode=g++ )

Add `--driver-mode=g++` as private compile/link option to each target.

CongMa13 (dismissed):
lgtm

cgmillette (dismissed):

cgmillette (approved):

PR #194, opened 2024-03-10
Use compressed debug symbols

CongMa13:
The size of file will be greater than 2G in Debug and RelWithDebug mode.

The use of compressed debug symbols significantly reduces library size, avoiding the need to remove targets in the address sanitizer build. The non-address sanitizer debug builds didn't drop GPU targets, but shared the same library size problem as the address sanitizer build. This change will also fix those builds.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -44,6 +44,12 @@ function(add_hiptensor_component COMPONENT_NAME COMPONENT_SOURCES)
     add_library(${COMPONENT_NAME} OBJECT ${COMPONENT_SOURCES})
     set_target_properties(${COMPONENT_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
     target_link_libraries(${COMPONENT_NAME} PRIVATE hip::device)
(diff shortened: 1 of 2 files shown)


mkarunan (dismissed):
LGTM

library/src/CMakeLists.txt
```diff
@@ -80,6 +86,12 @@ add_library(hiptensor SHARED
 add_library(hiptensor::hiptensor ALIAS hiptensor)
 
 target_compile_options(hiptensor PRIVATE ${CMAKE_CXX_FLAGS})
+if(HIPTENSOR_BUILD_COMPRESSED_DBG)
```
cgmillette, line 89: Do we need to carry this through the interface? Not sure if we want users to require this as well?

CongMa13: tested and PRIVATE worked.

Large obj file are from these contraction instance cpp. So -gz should not be inherited.

cgmillette (approved):

dlangbe (approved):

PR #195, opened 2024-03-10
Read the yaml config file as non-text

CongMa13:
There is no carriage linefeed at the end of each line of yaml file.

With the new llvm, should read the yaml file with default value `isText=false`

### DIFF
--- a/test/00_unit/yaml_test.cpp
+++ b/test/00_unit/yaml_test.cpp
@@ -127,6 +127,10 @@ int main(int argc, char* argv[])
     hiptensor::YamlConfigLoader<hiptensor::ContractionTestParams>::storeToFile(tmpFile, yee);
     auto yee1
         = hiptensor::YamlConfigLoader<hiptensor::ContractionTestParams>::loadFromFile(tmpFile);
+    if(!yee1)
+    {
+        return -1;
+    }
 
     return 0;
 }
--- a/test/llvm/CMakeLists.txt
+++ b/test/llvm/CMakeLists.txt
@@ -35,8 +35,13 @@ find_library(LLVMSupport_LIBRARY NAMES LLVMSupport PATHS ${LLVM_LIBRARY_DIR})
 
 message(STATUS "adding hiptensor component: hiptensor_llvm")
 message(STATUS "LLVM_LIBRARY_DIR: ${LLVM_LIBRARY_DIR}")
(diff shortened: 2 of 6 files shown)


cgmillette (approved):

test/00_unit/yaml_test.cpp
```diff
@@ -127,6 +127,10 @@ int main(int argc, char* argv[])
     hiptensor::YamlConfigLoader<hiptensor::ContractionTestParams>::storeToFile(tmpFile, yee);
     auto yee1
         = hiptensor::YamlConfigLoader<hiptensor::ContractionTestParams>::loadFromFile(tmpFile);
```
CongMa13, line 129: Found that the returned loading error was hidden by the unit test. Fixed it here.

test/llvm/CMakeLists.txt
```diff
@@ -35,9 +35,14 @@ find_library(LLVMSupport_LIBRARY NAMES LLVMSupport PATHS ${LLVM_LIBRARY_DIR})
 
 message(STATUS "adding hiptensor component: hiptensor_llvm")
 message(STATUS "LLVM_LIBRARY_DIR: ${LLVM_LIBRARY_DIR}")
+message(STATUS "LLVM_VERSION_MAJOR: ${LLVM_VERSION_MAJOR}")
 message(STATUS "LLVMObjectYAML_LIBRARY: ${LLVMObjectYAML_LIBRARY}")
 message(STATUS "LLVMSupport_LIBRARY: ${LLVMSupport_LIBRARY}")
```
CongMa13, line 41: Found that llvm::raw_fd_ostream(string & const filename, error_code ec) is only supported by llvm 7.0 and above

test/llvm/yaml_parser_impl.hpp
```diff
@@ -37,16 +37,16 @@ namespace hiptensor
 {
     template <typename ConfigT>
     /* static */
-    ConfigT YamlConfigLoader<ConfigT>::loadFromFile(std::string const& filePath)
+    std::optional<ConfigT> YamlConfigLoader<ConfigT>::loadFromFile(std::string const& filePath)
     {
         auto result = ConfigT{};
```
CongMa13, line 43: llvm 10.0 cannot parse the yaml file if the 2nd parameter is true.

PR #198, opened 2024-03-18
Only Searching for rocm-llvm-dev

CongMa13:
# Only searching for rocm-llvm-dev

HipTensor calls APIs of llvm package. But llvm is not installed with OS.

HipTensor should only use rocm-llvm-dev to reduce the 3rd party
dependency as rocm-llvm-dev is a part of rocm

Searching in /opt/rocm first

image

No `rocm-llvm-dev`, searching in `/usr/lib`. And stop if the found LLVM is not 7.0 and above. 
image

---------------

# Require version of LLVM > 6.0

In yaml_parser_impl.hpp, we call `raw_fd_ostream(filePaht, ec)`. 

This signature was added in LLVM 7. 

Now in a staging 1 docker, LLVM 6 and 10 are installed. CMake script will choose LLVM 6 by default.

Then we will get a building error about `raw_fd_ostream`. 

To fix this issue, cmake script should only find LLVM >= 7.0.

Unfortunately, the following line does not work. 

`find_package(LLVM 7.0 REQUIRED CONFIG)`

Though the cmake docs states that `7.0` means all version >= 7.0. LLVM does not support it. 

For LLVM, `find_package(LLVM 7.0 REQUIRED CONFIG)` only match `7.0.*`

 To make it more flexible, added an option `HIPTENSOR_LLVM_VERSION` with default value "". It means camke will use the first LLVM it finds by default. If the LLVM < 7.0. User need to specify a higher installed version .

### DIFF
--- a/test/llvm/CMakeLists.txt
+++ b/test/llvm/CMakeLists.txt
@@ -25,7 +25,12 @@
  ###############################################################################
 
 # Find / configure LLVM
-find_package(LLVM ${HIPTENSOR_LLVM_VERSION} REQUIRED CONFIG)
+find_package(LLVM CONFIG HINTS ${ROCM_DIR}/../../..)
+
+if (NOT LLVM_DIR)
+    message(FATAL_ERROR "HipTensor requires LLVM version 7.0 or higher,"
+        " which cannot be found. Please install rocm-llvm-dev package first." )
+endif()
 
 # Required:
 # LLVMObjectYAML for YAML parsing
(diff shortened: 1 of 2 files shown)


test/llvm/CMakeLists.txt
```diff
@@ -25,7 +25,11 @@
  ###############################################################################
 
 # Find / configure LLVM
-find_package(LLVM REQUIRED CONFIG)
```
CongMa13, line 28: 1. Removed REQUIRED to print both default and custom error message
2. HINTS => searching LLVM in `HINTS` before system path

cgmillette (dismissed):

dlangbe (dismissed):

cgmillette (approved):
LGTM

PR #203, opened 2024-03-25
Contraction extend ranks-M,N,K to 6

mkarunan:
Modifies to use ranks of M,N,K to 6. 
Remove Rank-2 instances for M,N,K and added Rank-6 scale and bilinear contraction instances.
Modified contraction solution to expand to lengths and strides of 6D.
Verified the existing 4d rank tensor contraction for brute force approach.
Disabled actor-critic tests as it needs retraining.
Modified CPU contraction implementation to use 6D rank variation.


cgmillette (approved):
Tested with CK's CI. Moving to merge

PR #208, opened 2024-04-10
Use custom env for RTD

samjwu:
Uses Doxygen 1.9.8 instead of 1.9.1, which has a bug that incorrectly fails the doc builds


samjwu:
doc build passes https://advanced-micro-devices-demo--208.com.readthedocs.build/projects/hipTensor/en/208/

cgmillette (approved):

PR #213, opened 2024-04-16
Make initArgs an private function

CongMa13:
The internal device memory needs to be freed after invoking contraction. To make the interface easier to use, the three steps: `initArgs`, `Run`,
   `release` are moved into `operator()`.  Users will never forget to
   release the memory.

### DIFF
--- a/library/src/contraction/contraction_cpu_reference_impl.hpp
+++ b/library/src/contraction/contraction_cpu_reference_impl.hpp
@@ -156,12 +156,12 @@ namespace hiptensor
                         indices.begin(), indices.end(), strides.begin(), std::size_t{0});
                 };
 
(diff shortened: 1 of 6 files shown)


library/src/contraction/contraction_cpu_reference_impl.hpp
```diff
@@ -156,12 +156,12 @@ namespace hiptensor
                         indices.begin(), indices.end(), strides.begin(), std::size_t{0});
                 };
 
-                if constexpr((std::is_same_v<ADataType, hipFloatComplex> &&
```
CongMa13, line 159: changed by clang-format

cgmillette (approved):

PR #211, opened 2024-04-15
Update docs

cgmillette:
hipTensor needs a docs overhaul to fit the new style, and hipTensor's new API and library dependencies:

* Update what-is-hiptensor
* Update installation.rst
* Update programmers-guide.rst
* Update api-reference.rst

Additional fixes include project configs and doxygen formatting in the code

PR #214, opened 2024-04-18
Fixing a few grammatical errors

SwRaw:


docs/index.rst
randyh62, line 1: I would change "Our documentation" to "The documentation", and I don't think there is a need for "what-is-hiptensor" when the paragraph could just appear in the index.

randyh62 (dismissed):

cgmillette:
Hi @SwRaw @randyh62 we are in the process of upgrading our documentation. Can you please review #211 and provide your feedbacks against that branch? There are a large amount of changes because we've updated the API and many functionalities.

PR #222, opened 2024-05-13
Added support to tensor reduction

CongMa13:
- Data type

    HIP_R_16F, HIP_R_16F, HIP_R_16F, HIPTENSOR_COMPUTE_16F,
    HIP_R_16F, HIP_R_16F, HIP_R_16F, HIPTENSOR_COMPUTE_32F,
    HIP_R_16BF, HIP_R_16BF, HIP_R_16BF, HIPTENSOR_COMPUTE_16BF,
    HIP_R_16BF, HIP_R_16BF, HIP_R_16BF, HIPTENSOR_COMPUTE_32F,
    HIP_R_32F, HIP_R_32F, HIP_R_32F, HIPTENSOR_COMPUTE_32F,
    HIP_R_64F, HIP_R_64F, HIP_R_64F, HIPTENSOR_COMPUTE_64F,

- Operator ADD MUL MIN MAX

Known Issues:
- CK is not supporting ADD/MUL/MIN/MAX with HIP_R_16F and HIP_R_16BF 
- CK is not supporting reduce the fastest changing extent.


library/include/hiptensor/hiptensor_types.hpp
```diff
@@ -104,10 +104,20 @@ typedef enum
     //! Identity operator
     HIPTENSOR_OP_IDENTITY = 1,
     //! Square root operator
-    HIPTENSOR_OP_SQRT = 3,
+    HIPTENSOR_OP_SQRT = 2,
```
cgmillette, line 107: Need to ensure these match the nv ones for api parity

library/src/reduction/hiptensor_reduction.cpp
```diff
@@ -0,0 +1,373 @@
...
+            }
+            if(!modeD)
+            {
+                printErrorMessage("modeD");
+            }
+            return errorCode;
+        }
+
+        const hiptensor::Hash            hashGenerator;
+        const std::unordered_set<size_t> supportedTypes = {
```
cgmillette, line 142: Nice. We should do this with permute, etc as well in a follow-up PR

cgmillette (approved):

library/include/hiptensor/hiptensor_types.hpp
CongMa13: Double checked. Value 2 is consistant with cuda. 

image

PR #227, opened 2024-05-27
Fix linting workflow and codeowner syntax

samjwu:
Update organization to ROCm for linting workflow

Use / in codeowner for pattern to include nested files and folders

RE: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
RE: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


cgmillette (approved):
LGTM. Thanks for the fix

PR #230, opened 2024-06-03
Reduce cpu reference

CongMa13:
1. Added reduce CPU reference. 
2. Split reduction_solution_instances.cpp to reduce the building time


library/src/reduction/reduction_solution_impl.hpp
```diff
@@ -100,20 +100,24 @@ namespace hiptensor
...
             }
 
-            std::array<ck::index_t, Traits::TensorRank>                              arrInLengths;
-            std::array<ck::index_t, Traits::TensorRank>                              arrInStrides;
-            std::array<ck::index_t, Traits::TensorRank - Traits::TensorNumReduceDim> arrOutLengths;
-            std::array<ck::index_t, Traits::TensorRank - Traits::TensorNumReduceDim> arrOutStrides;
-            std::array<ck::index_t, Traits::TensorNumReduceDim>                      reduceDims;
-            auto                                                                     toCKArr
+            constexpr ck::index_t OutputDim
+                = (Traits::TensorRank - Traits::TensorNumReduceDim)
```
cgmillette, line 105: Might need to adjust the handling of this due to ck::index_t is signed.
Suggest something like max(1, TensorRank-ReduceDim)

CongMa13: Added static_assert to ensure that Traits::TensorRank >= Traits::TensorNumReduceDim

PR #240, opened 2024-06-22
Add -mcmodel=large for ASAN pipeline

CongMa13:
Goal
Provide ASAN instrumented builds of all AMD built libraries
Improve release quality by running instrumented tests

Problem
hiptensor library is larger than 2GB if ASAN is enabled.

Solution
Add -mcmodel=large for ASAN pipeline

Check if CMake variable ADDRESS_SANITIZER is true
Check if env variable ADDRESS_SANITIZER is true
Check if CXXFLAGS contains "-fsanitize=address"
Tested with ASAN docker

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -60,6 +60,15 @@ include_directories(BEFORE
     ${PROJECT_SOURCE_DIR}/library/src/include
 )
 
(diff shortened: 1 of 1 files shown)


cgmillette (approved):
LGTM

dlangbe (approved):

PR #241, opened 2024-06-24
Reduce test

CongMa13:
Added unit tests of reduction:
* rank2-6

Found that some test cases could pass:
* when beta != 0
* when reducing the fastest growing dim
* when datatype is bf16 or f16

Comment out these test cases. Waiting for CK team to fix them.


library/src/reduction/reduction_solution_bf16_bf16_instance.cpp
```diff
@@ -0,0 +1,70 @@
...
+ *
+ *******************************************************************************/
+
+#include "hiptensor/internal/types.hpp"
+#include "reduction_solution.hpp"
+#include "reduction_solution_instances.hpp"
+
+namespace hiptensor
+{
+    void ReductionSolutionInstances::genReductionSolutionF32F32Instances()
```
cgmillette: Naming of this should be bf16?

library/src/reduction/reduction_solution_f16_f16_instance.cpp
```diff
@@ -0,0 +1,69 @@
...
+ *
+ *******************************************************************************/
+
+#include "hiptensor/internal/types.hpp"
+#include "reduction_solution.hpp"
+#include "reduction_solution_instances.hpp"
+
+namespace hiptensor
+{
+    void ReductionSolutionInstances::genReductionSolutionF32F32Instances()
```
cgmillette: Naming of this should be f16?

cgmillette: Also this is the 4_2 case only. Are we waiting for CK to update the M6 kernel?

test/03_reduction/reduction_test.cpp
```diff
@@ -0,0 +1,419 @@
...
+        EXPECT_TRUE((lengths.size() > 1) && (lengths.size() <= 6));
+        EXPECT_TRUE((outputDims.size() >= 1) && (outputDims.size() <= 6));
+
+        EXPECT_TRUE((op == HIPTENSOR_OP_ADD) || (op == HIPTENSOR_OP_MUL) || (op == HIPTENSOR_OP_MAX)
+                    || (op == HIPTENSOR_OP_MIN));
+
+        EXPECT_EQ(testType.size(), 2); // HIP_R_16F or HIP_R_32F
+        auto acDataType      = testType[0];
+        auto computeDataType = convertToComputeType(testType[1]);
+        EXPECT_TRUE(
```
cgmillette, line 106: Reduction instances only include f32 and f64 for now. Shall we disable the F16 and BF16 tests?

cgmillette (commented):
In general, looks great. Just a few comments from above. Please look into the docs build failures before approval!

cgmillette:
This looks like it has changes from #230 as well?

cgmillette (changes requested):
Please check other comments and have a look at broken docs build

library/src/reduction/reduction_solution_bf16_bf16_instance.cpp
CongMa13: Right now, CK has not fully supported bf16 and f16. There are building error to build these instances. So this file is not in the CMakeLists.txt file. I will remove this file.

library/src/reduction/reduction_solution_f16_f16_instance.cpp
CongMa13: Right now, CK has not fully supported bf16 and f16. There are building error to build these instances. So this file is not in the CMakeLists.txt file. I will remove this file.

test/03_reduction/reduction_test.cpp
CongMa13: Right now, CK has not fully supported bf16 and f16. There are building error to build these instances. So we cannot test f16 and bf16 now.

CongMa13:
> This looks like it has changes from #230 as well?

yeah, this includes #230 . I will rebase this PR after #230 is merged.

cgmillette (approved):
Looks good

CongMa13:
> Please check other comments and have a look at broken docs build

docs have been built successfully

PR #251, opened 2024-07-24
New fill kernel and validation tolerances

dlangbe:
Implements the new fill kernel from [PR 236](https://github.com/ROCm/hipTensor/pull/236) with supporting changes to the validation thresholds used throughout the testing infrastructure. The new validation thresholds should guarantee that validating CK kernels will also validate when used within hipTensor.

### DIFF
--- a/test/01_contraction/contraction_test.cpp
+++ b/test/01_contraction/contraction_test.cpp
@@ -266,7 +266,6 @@ namespace hiptensor
                 fillLaunchKernel<_Float16>((_Float16*)resource->deviceB().get(), elementsB, seed);
                 if(CDataType == HIP_R_16F)
                 {
-                    printf("Filling C\n\n");
                     fillLaunchKernel<_Float16>((_Float16*)resource->deviceC().get(), elementsCD, seed + 1);
                 }
                 fillValLaunchKernel<_Float16>((_Float16*)resource->deviceD().get(),
@@ -666,7 +665,8 @@ namespace hiptensor
                                             size_t{1},
                                             std::multiplies<size_t>());
 
-            auto eps = getEpsilon(computeType);
+            auto eps = getEpsilon(computeType == HIPTENSOR_COMPUTE_64F ? HIPTENSOR_COMPUTE_64F
+                : HIPTENSOR_COMPUTE_32F);
             double tolerance = 2 * nelems_k * eps;
 
             // use the same default tolerance value as CK
--- a/test/01_contraction/contraction_test.cpp
+++ b/test/01_contraction/contraction_test.cpp
@@ -257,14 +257,17 @@ namespace hiptensor
             auto resource = getResource();
             resource->resizeStorage(lengths, elementBytes);
 
+            uint32_t seed = static_cast<uint32_t>(std::time(nullptr));
+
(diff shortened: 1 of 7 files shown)


cgmillette (commented):
Looks great so far. I had a discussion with @CongMa13 and I we should add back in a seed value because each thread's hash value is deterministic. We don't want fill values to be the same for same-sized tensors. We can differentiate them at least by seeding different values for different tensors.

test/01_contraction/contraction_test.cpp
```diff
@@ -257,14 +257,17 @@ namespace hiptensor
             auto resource = getResource();
             resource->resizeStorage(lengths, elementBytes);
 
+            uint32_t seed = static_cast<uint32_t>(std::time(nullptr));
```
CongMa13: We cannot reproduce the input data if seed is changing every time the test runs. Then we cannot reproduce a bug if there is. 

So we can set a fix number as seed.

test/01_contraction/contraction_test.cpp
```diff
@@ -257,14 +257,17 @@ namespace hiptensor
...
             {
                 // Initialize matrix data on device
-                fillLaunchKernel<_Float16>((_Float16*)resource->deviceA().get(), elementsA);
-                fillLaunchKernel<_Float16>((_Float16*)resource->deviceB().get(), elementsB);
+                fillLaunchKernel<_Float16>((_Float16*)resource->deviceA().get(), elementsA, seed - 1);
+                fillLaunchKernel<_Float16>((_Float16*)resource->deviceB().get(), elementsB, seed);
                 if(CDataType == HIP_R_16F)
                 {
-                    fillLaunchKernel<_Float16>((_Float16*)resource->deviceC().get(), elementsCD);
+                    printf("Filling C\n\n");
```
CongMa13: Can we remove the printf?  :-)

test/01_contraction/contraction_test.cpp
```diff
@@ -656,13 +659,36 @@ namespace hiptensor
...
             resource->copyData(reference, resource->hostD(), sizeD);
 
+            // Compute tolerance based on compute type
+            auto dimension = a_ms_ks.mLengths.size() / 2;
+            auto nelems_k = std::accumulate(a_ms_ks.mLengths.begin() + dimension,
+                                            a_ms_ks.mLengths.end(),
+                                            size_t{1},
+                                            std::multiplies<size_t>());
+
+            auto eps = getEpsilon(computeType);
```
CongMa13: I noticed that CK only use epsilon of f32 or f64 to generate the tolerance. 
https://github.com/ROCm/composable_kernel/blob/b9ba5b26769292827cd65d3be221b4b0bdb5d4b8/profiler/include/profiler/profile_contraction_impl.hpp#L142-L143

It makes sense to me since epsilon of bf16 (0.0078) is too large for this formular. 

double tolerance = 2 * nelems_k * eps  

For example, when nelems_k  == 1000, tolerance is 7.8

We can follow the same rule as CK. 
`auto eps = getEpsilon(computeType == HIPTENSOR_COMPUTE_64F ? HIPTENSOR_COMPUTE_64F : HIPTENSOR_COMPUTE_32F);`

CongMa13 (approved):
lgtm

cgmillette (approved):
LGTM, all tests pass

PR #253, opened 2024-07-26
Fix for permutation cpu reference impl

jakpiase:
Fix for permutation cpu reference impl needed in https://github.com/ROCm/hipTensor/pull/251. Current reference implementation doesn't work correctly in some of cases and is a bit overengineered so I have made a new, simpler one which gives proper results. I have tested it on https://github.com/ROCm/hipTensor/pull/251 branch and both rank5 and rank6 permutation tests are passing with it.


CongMa13 (approved):
Awesome. Thank you for the fix. 

The code is cleaner than before.

cgmillette (approved):
LGTM

PR #261, opened 2024-08-07
Refuce f16

CongMa13:
- Support f16/f16, bf16/bf16, f16/f32, bf16/f32 reduction
- Support independent C and D
- Add another CK instance with InSrcVectorDim==1 which supports far right dim
- Support reducing a tensor to a scalar
- Support permutation of output of reduction
- Fix bugs


test/03_reduction/reduction_resource.cpp
```diff
@@ -86,31 +86,25 @@ namespace hiptensor
...
         mCurrentMatrixCElement = requiredElementCountC;
 
         mCurrentDataType = dataType;
 
+        const uint32_t seedA = 256;
+        const uint32_t seedC = 256;
         if(needFillDataA)
         {
-            fillRand(hostA(), deviceA(), getCurrentMatrixAMemorySize());
+            fillRand(hostA(), deviceA(), dataType, getCurrentMatrixAElement(), seedA);
```
CongMa13: fixed a bug here. Should be element count, not memory size.

test/03_reduction/reduction_resource.cpp
```diff
@@ -86,31 +86,25 @@ namespace hiptensor
...
         }
 
-        if(needFillDataC)
-        {
-            fillRand(hostC(), deviceC(), getCurrentMatrixCMemorySize());
-        }
+        fillRand(hostC(), deviceC(), dataType, getCurrentMatrixCElement(), seedC);
```
CongMa13: fixed another bug here. Input C should be reset for each test case since it was overwritten in the last case. 

C = alpah * reduce(A) + beta * C

test/03_reduction/reduction_test.cpp
```diff
@@ -372,21 +386,29 @@ namespace hiptensor
             }
             else if(acDataType == HIP_R_32F)
             {
+                auto reducedSize
```
CongMa13: Like contraction, the tolerance = lengthK * eps

library/src/hiptensor.cpp
```diff
@@ -151,18 +151,17 @@ hiptensorStatus_t hiptensorInitTensorDescriptor(const hiptensorHandle_t*     han
         return HIPTENSOR_STATUS_NOT_INITIALIZED;
     }
 
-    if((lens == nullptr)
+    if((lens == nullptr && strides != nullptr)
```
CongMa13, line 154: Need to allow lens == nullptr in TensorDescriptor. 

TensorDescriptor with lens == `nullptr` represents a scalar. So we can reduce a tensor to a scalar.  Like [[1,2,3], [1,2,3]] ---sum reduce--> 12.

library/src/reduction/reduction_solution_impl.hpp
```diff
@@ -128,18 +129,24 @@ namespace hiptensor
...
-            // @todo
-            // toCKArr( a_strides.empty() ? hiptensor::stridesFromLengths(a_lengths, HIPTENSOR_DATA_LAYOUT_COL_MAJOR) : a_strides, arrInStrides);
-            toCKArr(c_lengths, arrOutLengths);
-            toCKArr(c_strides.empty() ? hiptensor::stridesFromLengths(c_lengths) : c_strides,
+
+            auto ckCLengths = c_lengths;
+            auto ckCStrides = c_strides;
+            if(ckCLengths.empty())
+            {
+                ckCLengths.push_back(1);
```
CongMa13, line 137: Reduce a tensor to a scalar, len and stride are both 1

library/src/hiptensor.cpp
```diff
@@ -151,18 +151,17 @@ hiptensorStatus_t hiptensorInitTensorDescriptor(const hiptensorHandle_t*     han
...
        || ((dataType != HIP_R_16F) && (dataType != HIP_R_16BF) && (dataType != HIP_R_32F)
-           && (dataType != HIP_R_64F) && (dataType != HIP_C_32F)
-           && (dataType != HIP_C_64F))
+           && (dataType != HIP_R_64F) && (dataType != HIP_C_32F) && (dataType != HIP_C_64F))
        || ((unaryOp != HIPTENSOR_OP_IDENTITY) && (unaryOp != HIPTENSOR_OP_SQRT)))
     {
         auto errorCode = HIPTENSOR_STATUS_INVALID_VALUE;
-        if(lens == nullptr)
+        if(lens == nullptr && strides != nullptr)
```
CongMa13, line 160: | | lens| strides|
|-|-|-|
|values| null | null|
|values| non-null | null|
|values| non-null | non-null|

These values are fine.

library/src/reduction/hiptensor_reduction.cpp
```diff
@@ -111,18 +110,10 @@ namespace
             {
                 printErrorMessage("beta");
             }
-            if(!C)
```
CongMa13, line 114: C could be null if beta is 0

library/src/reduction/hiptensor_reduction.cpp
```diff
@@ -111,18 +110,10 @@ namespace
...
-            {
-                printErrorMessage("C");
-            }
             if(!descC)
             {
                 printErrorMessage("descC");
             }
-            if(!modeC)
```
CongMa13, line 122: modeC could be null if output is a scalar

library/src/reduction/reduction_solution_impl.hpp
```diff
@@ -214,24 +221,58 @@ namespace hiptensor
                                                                           PropagateNan,
                                                                           OutputIndex>;
 
-        std::vector<DeviceOpPtr> opPtrs;
-        ck::tensor_operation::device::instance::add_device_reduce_instance_blockwise<
-            InDataType,
```
CongMa13, line 219: `add_device_reduce_instance_blockwise` creates some instances that generate wrong output. 

Now we only use two instances that we trust.

test/03_reduction/reduction_test.hpp
```diff
@@ -43,7 +43,7 @@ namespace hiptensor
     using ReductionTestParams_t = std::tuple<typename ReductionTestParams::TestTypesT,
                                              typename ReductionTestParams::LogLevelT,
                                              typename ReductionTestParams::LengthsT,
-                                             typename ReductionTestParams::ReducedDimsT,
+                                             typename ReductionTestParams::OutputDimsT,
```
CongMa13, line 46: reducedDims is misleading name here. Should be outputDims.

test/03_reduction/reduction_test.cpp
```diff
@@ -279,31 +280,43 @@ namespace hiptensor
...
-                extentC.push_back(extent[mode]);
+            std::vector<int64_t> extentA(lengths.cbegin(), lengths.cend());
+            std::vector<int64_t> extentC(sortedOutputDims.cbegin(), sortedOutputDims.cend());
+            std::transform(extentC.cbegin(), extentC.cend(), extentC.begin(), [&lengths](auto dim) {
+                return lengths[dim];
+            });
             std::vector<int64_t> extentD(extentC);
 
+            std::vector<int64_t> strideD = hiptensor::stridesFromLengths(extentD);
+            if(!std::equal(outputDims.cbegin(), outputDims.cend(), sortedOutputDims.cbegin()))
```
CongMa13, line 276: support permutation of output

library/src/reduction/hiptensor_reduction.cpp
```diff
@@ -131,10 +122,6 @@ namespace
             {
                 printErrorMessage("descD");
             }
-            if(!modeD)
```
CongMa13, line 134: modeD is null if reducing to a scalar

test/03_reduction/reduction_resource.cpp
```diff
@@ -38,27 +38,31 @@ namespace hiptensor
...
+        , mDeviceD(Base::allocDevice(0))
         , mHostA(Base::allocHost(0))
         , mHostC(Base::allocHost(0))
+        , mHostD(Base::allocHost(0))
         , mCurrentDataType(HIP_R_32F)
-        , mCurrentMatrixAElement(0)
-        , mCurrentAllocByteA(0)
-        , mCurrentMatrixCElement(0)
-        , mCurrentAllocByteC(0)
+        , mCurrentInputElementCount(0)
```
CongMa13, line 46: rename variables

library/src/hiptensor.cpp
cgmillette: Isn't a scalar lengths = [1] ?

CongMa13: In CUDA, the lengths parameter is ignored when creating a tensor descriptor if the rank is 0. 
image

When I tested, set `nmodeC == 0` and `extentC={10}` or `extentC={3,2}`, the output is always correct. So extentC is ignored. 

We do the same thing. 

When `nmodeC ==0`, `lenghts` is empty in a tensor descriptor.

test/03_reduction/reduction_test.cpp
```diff
@@ -56,8 +57,8 @@ namespace hiptensor
     bool ReductionTest::checkDevice(hipDataType            datatype,
                                     hiptensorComputeType_t computeDataType) const
     {
-        return ((datatype == HIP_R_32F || computeDataType == HIP_R_32F) && isF32Supported())
-               || ((datatype == HIP_R_64F || computeDataType == HIP_R_64F) && isF64Supported());
+        return !(((datatype == HIP_R_32F || computeDataType == HIP_R_32F) && !isF32Supported())
+                 || ((datatype == HIP_R_64F || computeDataType == HIP_R_64F) && !isF64Supported()));
```
cgmillette, line 88: Need some f16 check here?

test/03_reduction/reduction_test.cpp
```diff
@@ -94,19 +95,21 @@ namespace hiptensor
         auto beta       = std::get<5>(param);
         auto op         = std::get<6>(param);
 
-        EXPECT_TRUE((lengths.size() > 1) && (lengths.size() <= 6));
-        EXPECT_TRUE((outputDims.size() >= 1) && (outputDims.size() <= 6));
+        EXPECT_TRUE((lengths.size() > 0) && (lengths.size() <= 6));
+        EXPECT_TRUE((outputDims.size() >= 0) && (outputDims.size() < 6));
```
cgmillette, line 126: For the case where outputDims == 6, possible that the operation is simply a permutation?
Should we consider this case?

test/03_reduction/reduction_test.cpp
```diff
@@ -160,9 +176,14 @@ namespace hiptensor
                         stream, (bfloat16_t*)resource->hostC().get(), elementsC);
                     stream << std::endl;
 
-                    stream << "Refenrence elements (" << elementsC << "):\n";
+                    stream << "Tensor D elements (" << elementsD << "):\n";
```
cgmillette: Can we factor this function away into a templated helper? Lots of repeated code for each dataType...
```
auto print_array_elements= [](auto type_tag, ...) {
    using DataT = typename type_tag::type;
    stream << "Tensor A elements (" << elementsA << "):\n";
                    hiptensorPrintArrayElements<DataT >(
                        stream, (DataT *)resource->hostA().get(), elementsA);
                    stream << std::endl;;
...
};

template <typename DataT> struct type_tag {using type = DataT;};

// Usage:
print_array_elements(type_tag<bfloat16_t>{}, ...)
```

test/03_reduction/reduction_test.cpp
```diff
@@ -296,9 +369,9 @@ namespace hiptensor
                                                                      resource->deviceC().get(),
                                                                      &descC,
                                                                      modeC.data(),
-                                                                     resource->deviceC().get(),
-                                                                     &descC,
-                                                                     modeC.data(),
+                                                                     resource->deviceD().get(),
+                                                                     &descD,
+                                                                     modeD.data(),
```
cgmillette, line 333: Memory leak below if you don't release the work pointer

library/src/reduction/reduction_cpu_reference.cpp
```diff
@@ -69,17 +76,27 @@ hiptensorStatus_t hiptensorReductionReference(const void*
...
     }
 
+    if(C && C != D)
+    {
+        // CK API can only process $D = alpha * reduce(A) + beta * D$
+        // Need to copy C to D if C != D
+        CHECK_HIP_ERROR(hipMemcpy(D,
+                                  C,
+                                  hiptensor::elementsFromLengths(descC->mLengths)
+                                      * hiptensor::hipDataTypeSize(descC->mType),
```
cgmillette, line 86: This needs to be released somewhere

library/src/reduction/reduction_cpu_reference_instances.cpp
```diff
@@ -28,104 +28,130 @@
...
-                                                  true,              \
-                                                  false>());         \
-    registerSolutions(enumerateReferenceSolutions<type,              \
-                                                  type,              \
-                                                  type,              \
-                                                  dim_count,         \
-                                                  reduced_dim_count, \
-                                                  HIPTENSOR_OP_MAX,  \
-                                                  true,              \
+#define REG_CPU_SOLUTION(dim_count, reduced_dim_count, type, computeType) \
```
cgmillette, line 31: rename reduced_dim_count to output_... as below

library/src/reduction/reduction_solution_5_4_f16_f32_instance.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "reduction_solution.hpp"
+#include "reduction_solution_instances.hpp"
+
+namespace hiptensor
+{
+    void ReductionSolutionInstances::genReductionSolution5x4F16F32Instances()
+    {
+
+        // add entries to mSolutionQuery
+        registerSolutions(enumerateReductionSolutions<ck::half_t,
```
cgmillette: Maybe we could wrap these up into a registration macro like you did with the reference solutions

library/src/reduction/reduction_meta_traits.hpp
```diff
@@ -61,9 +62,23 @@ namespace hiptensor
         constexpr static bool        TensorPropagateNan = PropagateNan;
```
cgmillette, line 62: NumReduceDim to output... ?

test/03_reduction/configs/rank1_test_params.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - [ HIP_R_16BF, HIP_R_16BF]
+  - [ HIP_R_16BF, HIP_R_32F]
+  - [ HIP_R_64F, HIP_R_64F]
+  - [ HIP_R_32F, HIP_R_32F]
+Alphas:
+  - 2.3
+Betas:
+  - 0.0
+  # - 2.3
+Lengths:
```
cgmillette, line 15: Be sure to enable non-zero beta when CK bug is fixed.

test/03_reduction/reduction_test.cpp
cgmillette: Can we allocate with the HipResource and use the unique_ptr manage the work memory here?

library/src/reduction/reduction_cpu_reference_instances.cpp
CongMa13: It is reduced dim count here. 

We use both reduced dim count and output dim count in hiptensor since CK's APIs use reduced dim count and cuTensor's APIs use output dim count. 

Output dim count is used in test cases since cuTensor APIs are called in test cases. 

reduced dim count is used internally within hiptensor.

library/src/reduction/reduction_meta_traits.hpp
CongMa13: We used reduced dim count internally since CK's APIs use reduced dim count.

test/03_reduction/reduction_test.cpp
CongMa13: Yeah, you are right. outputDims == 6 should be supported. 

I tested the current CK reduction API, it does not support outputDims == 6 right now. 

And the current hiptensor permutation API only supports `D = alpha * permute(A)`. We need to implement the extra summing step `D = alpha * permute(A) + beta * C`

I think we can use CK's elementwise op instances to do it. But need to more tests about it. 

It's not a trivial change. So it is better to add it in another ticket.  https://ontrack-internal.amd.com/browse/LWPHIPTENS-331

test/03_reduction/reduction_test.cpp
CongMa13: Fixed. Use HipResource to manage workspace buffer.

test/03_reduction/reduction_test.cpp
CongMa13: fixed. Added `printReductionTestInputOutput`

test/03_reduction/reduction_test.cpp
CongMa13: Is f16 supported on all platforms? 
I'm trying to skip f32, f64 test cases on platforms that do not support them.

test/03_reduction/configs/rank1_test_params.yaml
CongMa13: created a task to track it. https://ontrack-internal.amd.com/browse/LWPHIPTENS-330

library/src/reduction/reduction_solution_5_4_f16_f32_instance.cpp
CongMa13: Fixed.

cgmillette (approved):
LGTM pending CI

PR #266, opened 2024-08-20
Cherry pick Reduce f16 (#261) to release-staging/rocm-rel-6.3

CongMa13:
- Support f16/f16, bf16/bf16, f16/f32, bf16/f32 reduction
- Use f32 as compute type for f16/f16, bf16/bf16
- Support independent C and D
- Add another CK instance with InSrcVectorDim==1 which supports far right dim
- Support reducing a tensor to a scalar
- Support permutation of output of reduction
- Rename label [Reduced Dims] to [Output Dims] in reduction test config file 
- Commnet out all test cases that beta != 0. Will add the test cases back when the CK fix is merged into amd_master
- Fix bugs


cgmillette (approved):
LGTM pending CI

PR #276, opened 2024-10-07
HipTensor Benchmarking Overhaul

dlangbe:
Includes changes from [271](https://github.com/ROCm/hipTensor/pull/271) and [273](https://github.com/ROCm/hipTensor/pull/273) with additional fixes for printout formatting and controlled header printing.


library/src/hiptensor_options.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    void HiptensorOptions::setDefaultParams(bool val)
+    {
+        mUsingDefaultParams = val;
+    }
+
+    void HiptensorOptions::setValidation(std::string val)
+    {
+        if(val.compare("ON") == 0)
```
CongMa13, line 99: Could we convert `val` to capital before comparing with `std::toupper`? Then `val` could be `On`, `on` or `ON`

library/src/hiptensor_options.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    void HiptensorOptions::setOstream(std::string file)
+    {
+        mOstream.initializeStream(file);
+    }
+
+    void HiptensorOptions::setOmits(int mask)
+    {
+        if(mask & 1)
+            mOmitSkipped = true;
```
CongMa13, line 56: Should we set `mOmitSkipped` to false if `mask & 1 == 0`. 

Same for other options.

scripts/performance/BenchmarkPermutation.sh
```diff
@@ -0,0 +1,44 @@
...
+# Copyright (C) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+
+set -eux
+
+# ensure this script is in the cwd
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+output_dir=hiptensor-benchmarks
+build_dir=../../build/bin/
+config_dir=../../test/02_permutation/configs
```
CongMa13: Just wondered if this should be `..../configs/bench`?

test/03_reduction/reduction_test.cpp
```diff
@@ -359,94 +447,134 @@ namespace hiptensor
...
+                                           hipDataTypeSize(acDataType),
+                                           std::multiplies<size_t>());
+
+            size_t sizeCD = std::accumulate(extentC.begin(),
+                                            extentC.end(),
+                                            hipDataTypeSize(acDataType),
+                                            std::multiplies<size_t>());
+
+            mElapsedTimeMs        = float64_t(timeMs);
+            mTotalGFlops          = 2.0 * ((sizeA * sizeCD) / hipDataTypeSize(acDataType));
```
CongMa13: The flops of reduction is `sizeA / hipDataTypeSize(acDataType))`

Say there is a tensor A of shape [3,4,5], and the output B of reduction is [3,4]

The size of B is `3*4` and each entry is the summation of 5 entries of A. 

So, the summation is called `3*4*5` times.

library/src/hiptensor_options.cpp
cgmillette: Agreed.

library/src/include/hiptensor_options.hpp
```diff
@@ -49,16 +49,27 @@ namespace hiptensor
...
         ~HiptensorOptions()                  = default;
 
-        void parseOptions(int argc, char** argv);
+        void setOstream(std::string file);
         void setOmits(int mask);
+        void setDefaultParams(bool val);
+        void setValidation(std::string val);
+        void setHotRuns(int runs);
+        void setColdRuns(int runs);
+        void setInputFilename(std::string file);
```
cgmillette: We may want to be a bit more specific about what the files are. 
Suggest setInputYAMLFilename..
Suggest setOutputStreamFilename..

test/01_contraction/contraction_test.cpp
```diff
@@ -670,102 +797,153 @@ namespace hiptensor
...
+            CHECK_HIP_ERROR(hipEventSynchronize(stopEvent))
 
-            // Compute tolerance based on compute type
-            auto dimension = a_ms_ks.mLengths.size() / 2;
-            auto nelems_k  = std::accumulate(a_ms_ks.mLengths.begin() + dimension,
-                                            a_ms_ks.mLengths.end(),
-                                            size_t{1},
-                                            std::multiplies<size_t>());
+            auto timeMs = 0.0f;
+            CHECK_HIP_ERROR(hipEventElapsedTime(&timeMs, startEvent, stopEvent));
```
cgmillette, line 804: Don't forget to destroy your events

cgmillette (commented):
Changes are looking great so far. Should be good after the feedbacks are addressed

library/src/hiptensor_options.cpp
dlangbe: All omit flags are initialized to false

cgmillette: Yes, however if you were ever to call this function again, the non-masked items need to be reset back to false.

cgmillette (dismissed):
LGTM

cgmillette (approved):

PR #279, opened 2024-10-14
Column Major Contraction

dlangbe:
Default to using HIPTENSOR_DATA_LAYOUT_COL_MAJOR when constructing strides in `hiptensorInitTensorDescriptor` when no strides are specified by user. HIPTENSOR_DATA_LAYOUT_COL_MAJOR can be specified by user when configuring CMake project.


CongMa13 (dismissed):
lgtm.

cgmillette (dismissed):
We can get the fix in now, but must remember to cherry pick for 6.3 with the A/C run.

library/src/include/util.hpp
```diff
@@ -74,6 +74,27 @@ namespace hiptensor
     {
         return std::accumulate(lengths.begin(), lengths.end(), T{1}, std::multiplies<T>());
     }
+
+    const uint32_t HIPTENSOR_MAX_RANK = 6;
+
+    static inline uint32_t getRank(std::vector<std::size_t> const& a_ms_ks_strides)
```
CongMa13, line 88: Could you explain what this function does by an example?

cgmillette (dismissed):
LGTM

CongMa13 (approved):

PR #284, opened 2024-10-24
Add compiling flag --offload-compress

CongMa13:
The the size of binary files is much smaller with this flag.

The size of libhiptensor.so shrinks from 215M to 145M. (Release and only for gfx90a)

See https://github.com/ROCm/rocBLAS-internal/pull/2707

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
  #
  # MIT License
  #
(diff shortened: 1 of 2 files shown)


cgmillette (approved):
lgtm

dlangbe (approved):

PR #288, opened 2024-11-05
Add unit tests back

CongMa13:
CK fixed an issue that prevented it from supporting unit tests with beta is not equal to 0.

Add these unit tests back.


cgmillette (approved):
LGTM

PR #292, opened 2024-11-13
Sample Valgrind Hang Fix

dlangbe:
Modfied samples to use actor-critic selection as a temporary work around for Valgrind hang.


cgmillette (approved):
LGTM

PR #297, opened 2024-11-25
Permutation selection

CongMa13:
Added the permutation instance selection.

- Removed duplicated code in device folder. 
- Removed Query which is not used in solution_registry
- Added selection of instances based on lengths of input tensor
    - selection function is hard code now. 

The work flow 

image

PR #308, opened 2024-12-04
Permutation selection v2

CongMa13:
Added the permutation instance selection.

- Removed duplicated code in device folder. 
- Removed Query which is not used in solution_registry
- Added selection of instances based on lengths of input tensor
    - selection function is hard code now. 

The work flow 

image


CongMa13:
It is odd that the building result of #297 is not consistent in CI. Seems like the cached data are not removed. 

Create a new PR as the same as #297 to test CI

cgmillette:
Copy of #297

PR #326, opened 2025-01-14
set minimum ROCm version

CongMa13:
The solution in https://github.com/ROCm/rocBLAS/blob/develop/CMakeLists.txt  stopped working in https://github.com/ROCm/hipTensor/pull/326/checks?check_run_id=35951788707 

 

The reason is that rocm-cmake trys to extract the rocm version `ROCM_PLATFORM_VERSION` from variable `ROCM_DIR`. 

`ROCM_DIR` is supposed to be something like `/opt/rocm-6.3.3/`. But this is not mandatory. For example, in GFX942 CI pipeline,  ROCM_DIR is '/agent/_work/1/rocm/share/rocm/cmake' (real path). 

 

So `ROCM_PLATFORM_VERSION` is not defined in GFX942 CI pipeline. 

------

The real goal of this ticket is to check if the compiler support `__hip_fp8_e5m2` and `__hip_fp8_e4m3`. 

Created a cmake macro to check if the detected ROCm supports them. 

The image below shows that cmake command fails with ROCm 6.2

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,13 +25,14 @@
  ###############################################################################
 
 cmake_minimum_required(VERSION 3.14)
+message(STATUS "CMake version: ${CMAKE_VERSION}")
 
 # NOTE: This has to be initialized before the project() command appears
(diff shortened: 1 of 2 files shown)


cgmillette (dismissed):
lgtm

cgmillette (approved):
LGTM

PR #331, opened 2025-02-04
Support dynamic unary operators in permutation

CongMa13:
## Changes in this PR

1. Move find_package(ck) to CMakeLists.txt in root folder since unit tests also need it.
2. Add all unary operator enum values supported by cutensor 1.7
3. Add a new struct HiptensorUnaryOp. It use table-driven method to dispatch the unary operation.
4. Use HiptensorUnaryOp to instantiate permutation instances.
5. Remove operator types from the key of permutation instance registry table.
6. Change `add_hiptensor_unit_test` to accept multiple source files
7. Remove all permutation unit test that operator of tensor descriptor B is not `IDENTITY` since `IDENTITY` is the only allowed value according to the cutensor document.

### dynamic operator (sqrt--square)
 ./bin/simple_permutation
[1298050][2025-02-05 16:05:55][hipTensor][Performance][hiptensorPermutation] KernelId: 4994285073568944272 KernelName: 3_256_128_128_16_16_0_1_8_8, 13.663 ms, 0.020 TFlops, 78.585 GB/s
Finished!

### static operator (sqrt--square)
./bin/simple_permutation
[1311508][2025-02-05 17:09:44][hipTensor][Performance][hiptensorPermutation] KernelId: 4994285073568944272 KernelName: 3_256_128_128_16_16_0_1_8_8, 13.618 ms, 0.020 TFlops, 78.846 GB/s
Finished!

**NOTE**: There is a bug in the develop branch now. The calculation for `HIPTENSOR_OP_SQRT` is square, not square root.  To make it comparable, hard code to use "Square" for both static and dynamic version.


library/src/hiptensor.cpp
```diff
@@ -155,7 +155,9 @@ hiptensorStatus_t hiptensorInitTensorDescriptor(const hiptensorHandle_t*     han
     if((lens == nullptr && strides != nullptr)
        || ((dataType != HIP_R_16F) && (dataType != HIP_R_16BF) && (dataType != HIP_R_32F)
            && (dataType != HIP_R_64F) && (dataType != HIP_C_32F) && (dataType != HIP_C_64F))
-       || ((unaryOp != HIPTENSOR_OP_IDENTITY) && (unaryOp != HIPTENSOR_OP_SQRT)))
+       || ((unaryOp == HIPTENSOR_OP_ADD) || (unaryOp == HIPTENSOR_OP_MUL)
```
CongMa13, line 158: should not be binary operators

library/src/include/hiptensor_unary_element_wise_operation.hpp
```diff
@@ -0,0 +1,236 @@
...
+                {
+                }
+                __host__ __device__ HiptensorUnaryOp(const HiptensorUnaryOp& dynamic_op) = default;
+                __host__            __device__ ~HiptensorUnaryOp()                       = default;
+                __host__ __device__ HiptensorUnaryOp& operator=(const HiptensorUnaryOp& other)
+                    = default;
+
+                __host__ __device__ void operator()(float& y, const float& x) const
+                {
+                    ops[op_type](y, x);
```
CongMa13, line 179: Table-driven method to implement dynamic op. No conditional operation here. Faster than `if..else`

library/src/CMakeLists.txt
```diff
@@ -72,7 +69,7 @@ endif()
 
 # Generates hiptensor_contraction and hiptensor_contraction_instances
 add_subdirectory(contraction)
-# Generates hiptensor_permutation and hiptensor_permutation_instances
+# Generates hiptensor_permutation
 add_subdirectory(permutation)
 # Generates hiptensor_reduction and hiptensor_reduction_instances
```
dlangbe:
```suggestion
# Generates hiptensor_reduction
```

library/src/data_types.cpp
```diff
@@ -361,6 +361,94 @@ namespace hiptensor
         {
             return "HIPTENSOR_OP_SQRT";
         }
+        else if(opType == HIPTENSOR_OP_RELU)
```
dlangbe: Can this decision tree be changed to a switch statement?

CMakeLists.txt
```diff
@@ -160,6 +160,9 @@ if(NOT WIN32 AND ${hip_VERSION_FLAT} GREATER 600241132)
   add_compile_options("SHELL: -mllvm -amdgpu-function-calls=false")
 endif()
 
+find_package( composable_kernel 1.0.0 REQUIRED PATHS $ENV{CK_DIR}/lib/cmake /opt/rocm /opt/rocm/ck COMPONENTS device_contraction_operations device_reduction_operations device_other_operations)
```
cgmillette, line 173: hipTensor historically relied on CK package for static headers, and only included in components that used CK as a backend in case we needed to switch it out ever. Not sure if we want to drag it into the whole project.

We want to ensure hipTensor libraries deployment doesn't depend on CK shared libraries. Is this still the case?

library/src/hiptensor.cpp
cgmillette: HIPTENSOR_OP_IDENTITY doesn't work as a pass-through for binary?

library/src/include/hiptensor_unary_element_wise_operation.hpp
```diff
@@ -0,0 +1,236 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#pragma once
```
cgmillette: Let's stick with the #ifndef guard style like the rest of the project

library/src/include/hiptensor_unary_element_wise_operation.hpp
cgmillette: Might be a good idea to inspect the asm for the storage location of the function pointer array. This is also shared across all threads, so we should also ensure it is put into a s_reg for performance reasons

test/00_unit/unary_op_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    int block_size = 64;
+    int grid_size  = (array_size + block_size - 1) / block_size;
+    hipLaunchKernelGGL(
+        unary_op_kernel, dim3(grid_size), dim3(block_size), 0, 0, d_x, d_y, array_size, op_type);
+
+    // Copy result back to host
+    CHECK_HIP_ERROR(hipMemcpy(array_y, d_y, array_bytes, hipMemcpyDeviceToHost));
+
+    // Compare result with reference
+    // bool success = std::abs(array_c[0] - array_ref[0]) < 100 * std::numeric_limits<float>::epsilon();
```
cgmillette: Can remove

test/02_permutation/configs/extended/rank2_test_params.yaml
cgmillette, line 1: Do we plan on adding other operators to the extended test cases?

test/02_permutation/CMakeLists.txt
```diff
@@ -23,16 +23,19 @@
...
- ${CMAKE_CURRENT_SOURCE_DIR}/permutation_test.cpp)
+ set(PermutationCommonSources
+     ${HIPTENSOR_COMMON_TEST_SOURCES}
+     ${CMAKE_CURRENT_SOURCE_DIR}/permutation_resource.cpp
+     ${CMAKE_CURRENT_SOURCE_DIR}/permutation_test.cpp
+     )
 
 #  tests
-set (PermutationTestSources ${PermutationCommonSources}
-${CMAKE_CURRENT_SOURCE_DIR}/permutation_column_major_test.cpp
```
cgmillette, line 32: Are we removing the col_major test? Maybe we should keep it for regression purposes?

library/src/permutation/permutation_solution_impl.hpp
```diff
@@ -61,20 +62,23 @@ namespace hiptensor
...
-                      void const*                     A,
-                      void*                           B,
-                      std::vector<std::size_t> const& a_lengths,
-                      std::vector<std::size_t> const& a_strides,
-                      const int32_t                   modeA[],
-                      std::vector<std::size_t> const& b_lengths,
-                      std::vector<std::size_t> const& b_strides,
-                      const int32_t                   modeB[],
-                      const hipDataType               typeScalar) override
+        bool initArgs(void const*                        alpha,
```
cgmillette: I believe this level of the interface we have gone below API level data structures like tensor descriptors. To uncouple the interface from API, maybe we should add the operators as an argument instead of tensor descriptors?

cgmillette (commented):
Overall, this is a very refreshing PR to reduce complexity of kernel instances through dynamic operators. Looking great so far. Some minor details captured in my other comments.

test/00_unit/unary_op_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+float unaryOpOnDeviceTest(float x, hiptensorOperator_t op_type)
+{
+    const int array_size  = 64;
+    const int array_bytes = array_size * sizeof(float);
+
+    // Host arrays
+    float array_x[array_size] = {0.0F};
+    float array_y[array_size] = {0.0F};
+
+    array_x[0] = x;
```
dlangbe, line 56: Should we initialize all values with x, and then compare all values instead of just the first?

test/00_unit/unary_op_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    x   = 3.3F;
+    ref = 1 / x; // y = 1 / 3.3
+    y   = unaryOpOnHostTest(x, HIPTENSOR_OP_RCP);
+    EXPECT_FLOAT_EQ(y, ref);
+    y = unaryOpOnDeviceTest(x, HIPTENSOR_OP_RCP);
+    EXPECT_FLOAT_EQ(y, ref);
+
+    x   = 3.3F;
+    ref = 0.964428810727F; // y = 1 / (1 + e^3.3)
```
dlangbe:
```suggestion
    ref = 0.964428810727F; // y = 1 / (1 + e^-3.3)
```

test/02_permutation/configs/emulation/smoke/rank3_test_params.yaml
```diff
@@ -9,7 +9,6 @@ Alphas:
 Lengths:
   - [ 5, 2, 4]
 Operators:
-  - [HIPTENSOR_OP_IDENTITY, HIPTENSOR_OP_SQRT]
```
dlangbe:
```suggestion
- [HIPTENSOR_OP_NEG, HIPTENSOR_OP_IDENTITY]
```

test/02_permutation/configs/emulation/smoke/rank5_test_params.yaml
```diff
@@ -9,7 +9,6 @@ Alphas:
 Lengths:
   - [ 5, 2, 3, 4, 5]
 Operators:
-  - [HIPTENSOR_OP_SQRT, HIPTENSOR_OP_SQRT]
 Permuted Dims:
```
dlangbe, line 13:
```suggestion
- [HIPTENSOR_OP_NEG, HIPTENSOR_OP_IDENTITY]
Permuted Dims:
```

CMakeLists.txt
CongMa13: yeah, hiptensor doesn't depend on CK libraries.  Only header files are needed.  `find_package` is just for header files paths. 

Moved this `find_package( composable_kernel...) ` up to root folder just because I found that it has been included in all of `./test/CMakeLists.txt`, `./samples/CMakeLists.txt` and `./library/src/CMakeLists.txt`

library/src/hiptensor.cpp
CongMa13: `Unary` here does not mean __hiptensor elementwise unary API__. It is just a function that accepts one argument, like `sqrt(x)`. 

The check line 158-160 just says `ADD`, `MUL`, `MIN` and `MAX` they are not `unary` since they require two arguments.

library/src/include/hiptensor_unary_element_wise_operation.hpp
CongMa13: definitely.  Copied from CK and forgot change it.

library/src/include/hiptensor_unary_element_wise_operation.hpp
CongMa13: static array is saved in .text segment of elf file.

test/02_permutation/CMakeLists.txt
CongMa13: `permutation_column_major_test.cpp` is actually an empty test file. 😃  
Should remove it. 
image

test/00_unit/unary_op_test.cpp
CongMa13: yeah, you are right. This is not consistent.  

Actually, we only need 1 element in the array. Changed the size to 1.

library/src/permutation/permutation_solution_impl.hpp
CongMa13: fixed.

library/src/data_types.cpp
CongMa13: changed.

test/02_permutation/configs/extended/rank2_test_params.yaml
CongMa13: All these unary ops are equal to each other for tensor unit tests.  For example,

If the input tensor is[0.9234234, 2.34232423, -1.34232, 3.3234], hiptensor APIs handle it as a tensor with 4 random floats.  

After applying `neg` or `sqrt` or `sin`, it is still a tensor of 4 random floats for hiptensor APIs. 

So only test `neg` here and the tests for all other new unary operators are put into test  `unary_op_test`

cgmillette (approved):
LGTM!

PR #334, opened 2025-02-18
YAML Benchmark Ranges

dlangbe:
Added the ability to generate test lengths from a given set of ranges in test YAML files. The expected format is:
```
Ranges:
  - [lowerBound, upperBound, stepSize]
  - [lowerBound, upperBound, stepSize, numLengths]
Random Ranges:
  - [lowerBound, upperBound, numLengths]
```
The generated lengths are appended to the Lengths parameter given in the YAML file.

Additional information and some examples are provided in the new sample YAML files found in `configs/<test_name>_sample_config.yaml`.


test/01_contraction/contraction_test_helpers.hpp
```diff
@@ -68,9 +68,119 @@ auto inline load_config_params()
...
+                      uint64_t x = 1,
+                      uint64_t y = 1,
+                      uint64_t z = 1)
+{
+    uint64_t size = 4 * pow(i * j * k * x * y * z, 2);
+
+    return (size >= minMemory) && (size <= maxMemory);
+}
+
+void generate_lengths(std::vector<std::vector<size_t>>               ranges,
```
CongMa13: const & ranges

test/01_contraction/contraction_test_helpers.hpp
```diff
@@ -68,9 +68,119 @@ auto inline load_config_params()
     return testParams;
 }
 
+bool checkMemoryLimit(uint64_t minMemory,
```
CongMa13: This is a camel style function name and the function name below is snake style. 

Better only use one style.  Not quite sure which we should use. We can discuss in next standup.

test/01_contraction/contraction_test_helpers.hpp
```diff
@@ -68,9 +68,119 @@ auto inline load_config_params()
...
+                                    {
+                                        lengths.push_back({{y, x, k, j, i, y, x, k, j, i},
+                                                           {y, x, k, j, i, y, x, k, j, i},
+                                                           {y, x, k, j, i, y, x, k, j, i}});
+                                    }
+                                    else if(rank == 6)
+                                    {
+                                        lengths.push_back({{z, y, x, k, j, i, z, y, x, k, j, i},
+                                                           {z, y, x, k, j, i, z, y, x, k, j, i},
+                                                           {z, y, x, k, j, i, z, y, x, k, j, i}});
```
CongMa13: This expression indicates that the test cases only cover a subspace of the entire input space, namely m0...m6 == n0...n6 == k0...k6. All cases where m0...m6 != n0...n6 != k0...k6 are not tested. Can we cover the entire input space with a limited number of test cases?

test/01_contraction/contraction_test_helpers.hpp
```diff
@@ -68,9 +68,119 @@ auto inline load_config_params()
...
+
+    return (size >= minMemory) && (size <= maxMemory);
+}
+
+void generate_lengths(std::vector<std::vector<size_t>>               ranges,
+                      std::vector<std::vector<std::vector<size_t>>>& lengths,
+                      uint32_t                                       rank)
+{
+    // Minimum and maximum memory footprint of each tensor
+    uint64_t minMemory = 64;
```
cgmillette: These are not quite indicative of memory - I would interpret them more of an element count limit.
Probably best to have them as static const as well

test/01_contraction/contraction_test_helpers.hpp
```diff
@@ -68,9 +68,119 @@ auto inline load_config_params()
...
+        uint64_t min_z = rank > 5 ? minVal : 1;
+
+        uint64_t max_i = rank > 0 ? maxVal : 1;
+        uint64_t max_j = rank > 1 ? maxVal : 1;
+        uint64_t max_k = rank > 2 ? maxVal : 1;
+        uint64_t max_x = rank > 3 ? maxVal : 1;
+        uint64_t max_y = rank > 4 ? maxVal : 1;
+        uint64_t max_z = rank > 5 ? maxVal : 1;
+
+        for(z = min_z; z <= max_z; z *= step)
```
cgmillette: Forward thinking for the future, this function will require an extra set of max / min values and nested loop for each added rank. If we ever support rank = 16 as an example, this will explode pretty quickly. Also will have to do the same for the permute / reduce helpers.

This problem could be broken down further and could be re-used for the other helpers as well.

Common theme here is that you are basically taking the range and producing two combinatorial products:
1. a combination range for each tensor size, e.g., rank1 {m,  k} = {n, k} = {m, n} = { comb[0 - N], comb[0 - N] }
2. a combination range for each tensor product e.g., rank1 {  {m, k}, {n, k}, {m, n} } =  { comb[ {m, k}, {n, k}, {m, n} ] }

TLDR: We may want to split this up into more basic combinatorial functions: 
- generateRange(lower, upper, step) // linear range generator
- combine(X, Y); // Combinatorial machine that combines each element of X with each element of Y
- combineRand(X, Y, limit) // Combinatorial machine that combines randomly each element of X with maximum <limit> elements of Y, to control the number of possibilities

test/01_contraction/contraction_test_helpers.hpp
cgmillette: Maybe we can use a randomization idea like suggested above

cgmillette:
@dlangbe  Test for gfx908 over time limit - how shall we address it?

test/hiptensor_length_generation.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+            size_t elementsB = std::accumulate(nLengths.begin(), nLengths.end(), size_t{1}, std::multiplies()) *
+                                std::accumulate(kLengths.begin(), kLengths.end(), size_t{1}, std::multiplies());
+
+            size_t elementsD = std::accumulate(mLengths.begin(), mLengths.end(), size_t{1}, std::multiplies()) *
+                                std::accumulate(nLengths.begin(), nLengths.end(), size_t{1}, std::multiplies());
+
+            elements = elementsA + elementsB + elementsD + elementsD;
+        }
+        std::cout << elements << std::endl;
```
CongMa13: Could we remove this line if it is for debugging purposes?

test/hiptensor_length_generation.hpp
```diff
@@ -0,0 +1,398 @@
...
+        size_t range = upper - lower;
+
+        // Seed the randomization
+        std::srand(0);
+
+        // Create a hash map for storing previously utilized size indices to prevent duplicates
+        std::unordered_map<std::vector<size_t>, int, Hash> map;
+
+        // Determine if the possible number of size combinations exceeds the given totalSizes and update accordingly
+        size_t possibleSizes = randomizeFromRange ? pow(range, rank) : pow(uniqueSizes, rank);
```
CongMa13, line 164: Not quite sure what `pow(range, rank)` is

`range = upper - lower`

if rank == 2

```
Ranges:
  - [2, 8196, 2]
```

 `pow(range, rank) == (8196 - 2)^2` . Wondering what this represents?

test/01_contraction/configs/bench/bilinear_test_params_rank1.yaml
```diff
@@ -14,20 +14,17 @@ Operators:
...
   - [0]
-  - [2]
 Lengths:
-  - [[128, 128], [128, 128], [128, 128]]
-  - [[256, 111], [64, 111], [256, 64]]
-  - [[139, 256], [141, 256], [139, 141]]
+  - [[2, 2], [2, 2], [2, 2]]
+Ranges:
+  - [2, 8196, 2]
+Random Ranges:
```
CongMa13, line 27: Should we add some data for `Random Ranges`?

test/hiptensor_length_generation.hpp
CongMa13: Explain by Chris. This is the size of the space of all testable points.

CongMa13 (dismissed):

cgmillette:
@dlangbe LGTM, just need to address the conflicts

library/src/include/hash.hpp
```diff
@@ -57,6 +58,17 @@ namespace hiptensor
...
 
+        template <typename T>
+        void operator()(std::size_t& seed, std::vector<T> const& vec) const
+        {
+            std::size_t tempSeed = 0;
+            for (const auto& element : vec)
+            {
+                operator()(tempSeed, element);
+                seed ^= tempSeed;
+            }
```
CongMa13:
```suggestion
            for (const auto& element : vec)
            {
                operator()(seed, element);
            }
```

CongMa13 (approved):
lgtm

PR #336, opened 2025-02-24
Remove `__device__ ck::math::cos`

CongMa13:
There was an issue in CK that `__device__ ck::math::cos` was missing.

Added the definition of `cos` function in hiptensor as a workaround of this issue. Remove the `cos` function in this commit since CK has added it.


cgmillette (dismissed):
LGTM

cgmillette (approved):
LGTM

PR #344, opened 2025-03-20
Enable rank2 elementwise trinary operation unit tests

CongMa13:
There was a crash issue in unit test rank2_elementwise_trinary_op_test. 
But it was fixed in 4b28f6f10acfeca0fba47b86667fa632edaa1b9c. 

So enable rank2_elementwise_trinary_op_test again.


cgmillette (approved):
lgtm

PR #358, opened 2025-04-26
Reduce unary

CongMa13:
CK does not support hiptensor's requirements. 

	1. CK does not support opC
	2. CK opA applies on padding elements which is not what we want. 

CK does not support opC

image

CK only supports opA (even it does not work well for hiptensor). 

We need to change the CK classes to add opC. 

Solution: 
	1. Copied some files from CK to reduce/device folder. 
	2. Added typename PriorDestElementwiseOperation to the interface of these classes. 
	3. Applied the op on elements in hiptensor_gridwise_2d_reduction_multiblock.hpp

CK opA does not work well for hiptensor

CK opA applies on padding elements which is not what we want. 

For example, 
the input is [1, 2 ,3], reduceOP is MAX, the in_element_op is NEG, the expected output is [-1].
But CK pads the input to [1, 2, 3, -inf, -inf, …, -inf], and apply NEG to padded array and get [-1, -2, -3, inf, …, inf]. 
The reduced result is `inf`.

solution:
	1. Copied library/src/reduction/device/hiptensor_dynamic_buffer.hpp  to reduce/device folder
	2. Added InElementOp to the interface of HiptensorDynamicBuffer
	3. Applied the op when calling Get()


library/src/reduction/device/hiptensor_device_reduce_multiblock.hpp
```diff
@@ -487,6 +500,7 @@ namespace ck
 
                 static bool IsSupportedArgument(const Argument* pArg)
                 {
+                    // The check was changed for LWPHIPTENS-371
```
cgmillette: Need to remove internal reference

library/src/reduction/device/hiptensor_dynamic_buffer.hpp
```diff
@@ -0,0 +1,530 @@
...
+#include "enable_if.hpp"
+#include "generic_memory_space_atomic.hpp"
+
+namespace ck
+{
+
+    // T may be scalar or vector
+    // X may be scalar or vector
+    // T and X have same scalar type
+    // X contains multiple T
```
cgmillette, line 42: What is X here?

CongMa13: X is function template parameter. See line 91.

library/src/reduction/device/hiptensor_device_reduce_multiblock.hpp
CongMa13: sure
```suggestion
                    // workaround for fixed InSrcVectorSize
```

cgmillette (approved):
LGTM

dlangbe (approved):

PR #361, opened 2025-05-08
Improve the code coverage of unit tests

CongMa13:
- remove unused code
- move functions used only in unit test to test folder
- add more unit tests

image


CongMa13:
[hiptensor.code_coverage.tar.gz](https://github.com/user-attachments/files/20106527/hiptensor.code_coverage.tar.gz)

library/src/hiptensor.cpp
```diff
@@ -466,22 +466,3 @@ hiptensorStatus_t hiptensorLoggerForceDisable()
     logger->disable();
     return HIPTENSOR_STATUS_SUCCESS;
 }
-
```
cgmillette: This is part of the API, which shouldn't be removed.

library/src/reduction/reduction_types.hpp
```diff
@@ -41,50 +41,14 @@ namespace hiptensor
                           || (opReduce == HIPTENSOR_OP_MIN) || (opReduce == HIPTENSOR_OP_MAX),
                       "opReduce is not supported");
 
-        constexpr auto reduceOpId = (opReduce == HIPTENSOR_OP_ADD)   ? ck::ReduceTensorOp::ADD
-                                    : (opReduce == HIPTENSOR_OP_MUL) ? ck::ReduceTensorOp::MUL
-                                    : (opReduce == HIPTENSOR_OP_MIN) ? ck::ReduceTensorOp::MIN
-                                                                     : ck::ReduceTensorOp::MAX;
+        constexpr auto reduceOpId = (opReduce == HIPTENSOR_OP_ADD)
```
cgmillette: To make this more extensible, you might want to throw it into a template class with overrides for each specific op mapping.

```
template<hiptensorOperator_t>
struct CKReduceOperator;

template<>
struct CKReduceOperator<HIPTENSOR_OP_ADD>
{ 
  static constexpr inline auto value() { return ck::ReduceTensorOp::ADD; }
};
...
```

CongMa13:
Removed some vulnerable test cases. Instead, change the reduction test config files to cover the same code path.

cgmillette (approved):
LGTM

PR #363, opened 2025-05-14
disable warning about 'amdgpu-waves-per-eu'

CongMa13:
Use the same compile option to disable warning about 'amdgpu-waves-per-eu'


cgmillette (approved):
lgtm

PR #364, opened 2025-05-16
fix rank5 permutation segfault

Ryker0627:
In function
   static std::vector<ck::index_t> findRepresentPointOfSubSpace(std::vector<std::size_t> const& lengths)

The array "edgeList" only has 5 elements. So when rank 5 tensor is used, the size of array "lengths" is 5, the code:
      auto&    edges = edgeList[lengths.size()];
will get an "edges" that is out of boundary of array "edgeList".
Then in code:
    auto lower_edge = len <= 32 ? 32 : *std::lower_bound(edges.cbegin(), edges.cend(), len);
 it will  cause memory crash when any dimension of the rank 5 tensor greater than 32 because "edges" passed into  "std::lower_bound" is not an effective array. 

The fix is to avoid calling "findRepresentPointOfSubSpace" function when rank 5 is used and also check  "lengths" size to make sure it is less than "edgeList" size.

### DIFF
--- a/library/src/permutation/permutation_instance_selection.cpp
+++ b/library/src/permutation/permutation_instance_selection.cpp
@@ -4103,6 +4103,7 @@ namespace hiptensor
             {32, 128, 512, 2048, 8192, 32768, 131072, 524288},
             {32, 256, 2048, 16384, 131072},
         };
+        if(lengths.size()>=edgeList.size()) return {};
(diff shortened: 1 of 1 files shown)


CongMa13:
The fix looks good to me.

cgmillette (approved):
lgtm

PR #393, opened 2025-08-25
Free allocated data in tests and clear Valgrind leak report

evedovelli:
## Motivation

Valgrind memory check has reported leaks when run with hipTensor, and these changes fix the leaks related to hipTensor code. All the found memory leaks were in test code only, so there are no impacts in the library or for anyone properly using the library APIs.

## Technical Details

Multiple tests were not calling the propre hipTensor APIs to detroy the created objects (plans, handlers, etc.). Also, some tests were using `test/hip_resource` to allocate host and device memory and this test library was not providing the APIs to free these allocated data.

## Test Plan

- All the tests were executed for regression verification.
- A small subset of the tests (2 tests), previously reporting memory leaks, were re-executed with valgrind.

## Test Result

- All tests are passing.
- All the *definitely lost* and *indirect lost* blocks have been cleared - no more leaks
  - Some few blocks are still marked as *possibly lost*, but the stack trace points to libhsa-runtime64.

## Submission Checklist

### DIFF
--- a/test/01_contraction/contraction_resource.cpp
+++ b/test/01_contraction/contraction_resource.cpp
@@ -66,6 +66,18 @@ namespace hiptensor
     {
     }
 
+    ContractionResource::~ContractionResource()
+    {
+        Base::freeDevice(mDeviceA);
+        Base::freeDevice(mDeviceB);
+        Base::freeDevice(mDeviceC);
+        Base::freeDevice(mDeviceD);
+        Base::freeHost(mHostA);
+        Base::freeHost(mHostB);
+        Base::freeHost(mHostC);
+        Base::freeHost(mHostD);
+    }
+
     void ContractionResource::copyHostToDeviceAll(ElementBytes const& bytesPerElement)
     {
         Base::copyData(mDeviceA,
--- a/test/01_contraction/contraction_resource.cpp
+++ b/test/01_contraction/contraction_resource.cpp
@@ -66,18 +66,6 @@ namespace hiptensor
     {
     }
 
-    ContractionResource::~ContractionResource()
-    {
-        Base::freeDevice(mDeviceA);
-        Base::freeDevice(mDeviceB);
-        Base::freeDevice(mDeviceC);
-        Base::freeDevice(mDeviceD);
-        Base::freeHost(mHostA);
-        Base::freeHost(mHostB);
-        Base::freeHost(mHostC);
-        Base::freeHost(mHostD);
-    }
-
     void ContractionResource::copyHostToDeviceAll(ElementBytes const& bytesPerElement)
     {
         Base::copyData(mDeviceA,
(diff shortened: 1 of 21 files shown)


test/01_contraction/contraction_resource.cpp
```diff
@@ -66,6 +66,18 @@ namespace hiptensor
     {
     }
 
+    ContractionResource::~ContractionResource()
+    {
+        Base::freeDevice(mDeviceA);
```
cgmillette: The host and device pointers are unique_ptr objects which are (should be) self-managed resources. Is this necessary?

test/01_contraction/contraction_test.cpp
```diff
@@ -711,6 +711,8 @@ namespace hiptensor
                         stream, (hipDoubleComplex*)references.get(), elementsCD);
                     stream << std::endl;
                 }
+
+                resource->freeHost(D);
```
cgmillette: Same thing here: D is a scoped unique_ptr. Is the de-allocator not being called automatically?

test/02_elementwise/elementwise_binary_op_test.cpp
```diff
@@ -568,20 +566,23 @@ namespace hiptensor
             } // if (testOptions->performValidation())
 
             CHECK_HIPTENSOR_ERROR(hiptensorDestroy(handle));
+            CHECK_HIPTENSOR_ERROR(hiptensorDestroyPlanPreference(planPref));
```
cgmillette, line 569: Good catch. These are because of the 2.0 API changes

test/02_elementwise/elementwise_binary_op_test.cpp
```diff
@@ -568,20 +566,23 @@ namespace hiptensor
...
+            CHECK_HIPTENSOR_ERROR(hiptensorDestroyPlan(plan));
```
cgmillette, line 570: +1 on the CHECK_HIPTENSOR_ERROR additions, nice catch

test/01_contraction/contraction_resource.cpp
evedovelli: Valgrind was reporting mem leaks associated with these objects, but I agree with you and I can see the unique_ptr have the proper deleters to free the allocated memory. Valgrind might have been reporting these leaks as "possibly lost".

I will run one of these tests and certify that the deleter is called and free the memory properly and undo these changes if confirmed.

evedovelli: I verified with rocgdb that the unique_ptr deleter gets called by the end of the test and that the memory gets freed. Valgrind was reporting these blocs as "possibly lost". So I've undone these cahnges.

test/02_elementwise/elementwise_binary_op_test.cpp
evedovelli: Reverted.

test/01_contraction/contraction_test.cpp
evedovelli: Reverted.

cgmillette (approved):
LGTM

PR #395, opened 2025-08-28
Fix Doxygen error in docs

yugang-amd:
Rename hiptensor-version.hpp.in to hiptensor-version.hpp to fix the following Doxygen error:

image

Note: This is a workaround for docs build so it only applies to the 7.0 docs branch.


alexxu-amd (approved):
LGTM

PR #397, opened 2025-09-13
Implement Plan Cache

Ryker0627:
## Motivation

Implement Plan Cache to hipTensor

## Technical Details
1. implement lookup_table<problem_id, solution_instance_id>:
The key is a hash id of the input problem. For example, hash id of a contraction problem is:  problem_id = hash({ Tag, DatatypeA, DatatypeB, DatatypeC, DatatypeD, DatatypeCompute, lengths, strides, modes}) 
Tag: This can be useful if the state of the hardware cache between these two calls is substantially different (i.e., affecting the measured runtime of the kernels).  
plan cache can evict cache entries using a least-recently-used (LRU) policy.
2. the cache is implemented in a thread-safe manner and is shared across all threads that use the same hiptensorHandle_t.  
3. plan cache table is a member of hiptensorHandle.
4. implement autotuning: users can automatically find the best implementation for the given problem and thereby increase performance. But the ranked solution system(from fastest to slowest) needs to be done in future.
5. Implement store/load the state of the cache to disk file.
6. Implement "export HIPTENSOR_DISABLE_PLAN_CACHE = ON/OFF" to disable or enable plan cache.
7. Implement API function like hiptensorPlanPreferenceSetAttribute to set plan cache related attributes.
8. Add sample fille "simple_contraction_plan_cache.cpp".

## Test Plan
All test cases and samples

## Test Result

All tests passed

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


library/src/contraction/contraction_selection.cpp
```diff
@@ -112,6 +115,10 @@ namespace hiptensor
             0,
         };
 
+        std::vector<float> sol_times(candidates.size(),1e100);
```
evedovelli: I believe you're using `1e100` just as a very large number here, right? If so, consider using `std::numeric_limits<float>::max()` instead.

library/src/contraction/contraction_selection.cpp
```diff
@@ -194,6 +205,14 @@ namespace hiptensor
 
         *winner = bestSolution;
 
+        //Sort candidates based on performance (from fastest to slowest)
+        std::sort(indices.begin(), indices.end(), [&](int i, int j) {
+           return sol_times[i] < sol_times[j];
+        });
+        std::vector<ContractionSolution*> tmpCandidates = candidates;
+        candidates.clear();
+        for(auto idx:indices) candidates.push_back(tmpCandidates[idx]);
```
evedovelli: You can use `std::move(tmpCandidates[idx])` to avoid copying each object.

library/src/contraction/hiptensor_contraction.cpp
```diff
@@ -298,18 +305,16 @@ hiptensorStatus_t hiptensorContract(const hiptensorHandle_t handle,
         return checkResult;
     }
 
-    auto realHandle = hiptensor::Handle::toHandle((int64_t*)handle->fields);
-
     // Ensure current HIP device is same as the handle.
     hiptensor::HipDevice currentDevice;
-    if((int)currentDevice.getDeviceId() != realHandle->getDevice().getDeviceId())
+    if((int)currentDevice.getDeviceId() != handle->mDevice.getDeviceId())
```
evedovelli: Wouldn't it be better calling `handle->getDevice()` to avoid accessing the private member `mDevice` directly? The same comment applies for the similar uses below in this file.

library/src/contraction/hiptensor_contraction.cpp
```diff
@@ -589,52 +592,69 @@ hiptensorStatus_t contractionInitPlan(const hiptensorHandle_t              handl
...
+                                                desc->mModeC,
+                                                EDataType,
+                                                hiptensor::getTensorLengths(desc->mDescD),
+                                                hiptensor::getTensorStrides(desc->mDescD),
+                                                desc->mModeD,
+                                                desc->mDescCompute,
+                                                workspaceSizeLimit);
+            //Save solutions (from fastest to slowest) for plan cache autotune
+            pref->mCandidates.clear();
+            for(auto candidate:candidates) pref->mCandidates.push_back(candidate);
```
evedovelli: Consider using `push_back(std::move(candidate))`

library/src/contraction/hiptensor_contraction.cpp
```diff
@@ -589,52 +592,69 @@ hiptensorStatus_t contractionInitPlan(const hiptensorHandle_t              handl
...
+                                                hiptensor::getTensorStrides(desc->mDescC),
+                                                desc->mModeC,
+                                                EDataType,
+                                                hiptensor::getTensorLengths(desc->mDescD),
+                                                hiptensor::getTensorStrides(desc->mDescD),
+                                                desc->mModeD,
+                                                desc->mDescCompute,
+                                                workspaceSizeLimit);
+            //Save solutions (from fastest to slowest) for plan cache autotune
+            pref->mCandidates.clear();
```
evedovelli, line 647: It's unfortunate we need to clear the `candidates` vector inside `bruteForceModel` and reinsert the sorted elements, and then we need to do the same for `pref->mCandidates` here. If somehow the `bruteForceModel` could process `pref->mCandidates` directly, we'd only need one clear + push_back loop.

library/src/hiptensor.cpp
```diff
@@ -462,8 +493,16 @@ hiptensorStatus_t hiptensorCreatePlanPreference(const hiptensorHandle_t    handl
                                                 hiptensorJitMode_t         jitMode)
 {
     *pref                        = new hiptensorPlanPreference();
-    (*pref)->mSelectionAlgorithm = algo;
+
+    (*pref)->mAutotuneMode       = HIPTENSOR_AUTOTUNE_MODE_NONE;
+    (*pref)->mCacheMode          = HIPTENSOR_CACHE_MODE_PEDANTIC;
+    (*pref)->mIncrementalCount   = 0;
+    (*pref)->mKernelrank         = 0;
```
evedovelli: I suggest renaming `mKernelrank` into `mKernelRank` to keep the camel case convention from other attributes.

library/src/include/plan_cache.hpp
```diff
@@ -0,0 +1,250 @@
...
+                solPtr = sol;
+                break;
+            }
+        }
+        return solPtr;
+    }
+
+    //Updatable priority queue
+    template <typename T, typename PriorityType, typename Compare = std::less<PriorityType>>
+    class Updatable_Priority_Queue
```
evedovelli: I recommend renaming this class as `UpdatablePriorityQueue` to follow the same naming convention as the other classes.

library/src/include/plan_cache.hpp
```diff
@@ -0,0 +1,250 @@
...
+        void push(const T& key, const PriorityType& priority)
+        {
+            Element *item = new Element{key, priority, heap_data.size()};
+            heap_data.push_back(item);
+            mp_key_to_element[key] = heap_data.back();
+            sift_up(heap_data.size() - 1);
+        }
+
+        //Update the priority of an existing element
+        bool update_item(const T& key, const PriorityType& new_priority)
```
evedovelli: Consider renaming this method and the others in this class to use camel case, so they follow the convention from other classes in hipTensor.

library/src/include/plan_cache.hpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        ~Updatable_Priority_Queue()
+        {
+            for(auto elem:heap_data) delete elem;
+            heap_data.clear();
+        }
+
+    private:
+        std::vector<Element*> heap_data;
```
evedovelli: I recommend naming the attributes following the convention mHeapData (m for member), which is the same convention used in other classes.

library/src/include/plan_cache.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+    private:
+        //Max size of the hash table
+        uint32_t max_cachelines;
+
+        //Hash table to store solution uid for lookup by hash key
+        std::unordered_map<HashId,Uid> mPlanCacheLines;
+
+        //Heap to sort the solution uid by used time
+        Updatable_Priority_Queue<HashId, time_point> pq_UidUsedTimes;
```
evedovelli: Can't we define `pq_UidUsedTimes` as `std::priority_queue` directly? So we wouldn't have to maintain our own implementation of a priority queue.

evedovelli: Also, consider renaming `pq_UidUsedTimes` as `mPqUidUsedTimes`

library/src/include/plan_cache.hpp
```diff
@@ -0,0 +1,250 @@
...
+        //Function for serialization to disk
+        hiptensorStatus_t writeFile(const char filename[]);
+
+        //Function for deserialization from disk
+        hiptensorStatus_t readFile(const char filename[]);
+
+        uint32_t getCachelinesNum() {return mPlanCacheLines.size();}
+
+       //Resize Plan Cache
+       void Resize(uint32_t numEntries);
```
evedovelli: The indentation is off by 1 column here.

There are multiple formatting inconsistences across the changes. I see this repository contains a .clang-format defined in the top directory. Could you run clang-format for all the files you modified in order to apply the formatting rules defined in that file?

library/src/include/plancache_autotune.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+namespace hiptensor
+{
+    class PlancacheAutotuneMgr : public LazySingleton<PlancacheAutotuneMgr>
+    {
+    private:
+
+        std::unordered_map<std::string,bool> mpAutotuneStarted; //If autotune process has been started
+        std::unordered_map<std::string,uint32_t> mpCallCount;  //Number of calling a function
+        std::unordered_map<std::string,std::pair<float,void*>> bestSolution; //The best solution pair (run time <--> solution)
```
evedovelli: Consider using an enum as the key for the maps above. It'd limit the options when inserting and searching for elements, avoiding misspelling errors. Besides, the enum comparison is faster than the string comparison when finding in the map.

evedovelli: I suggest renaming bestSolution as mBestSolution

evedovelli: What does the `mp` in `mpCallCount` and `mpAutotuneStarted` stands for?

library/src/include/plancache_autotune.hpp
```diff
@@ -0,0 +1,141 @@
...
+        PlancacheAutotuneMgr& operator=(PlancacheAutotuneMgr&&) = delete;
+
+    public:
+
+        friend std::unique_ptr<PlancacheAutotuneMgr> std::make_unique<PlancacheAutotuneMgr>();
+
+        //Start the autotune by setting initial status
+        void startAutotune(const std::string strFuncName)
+        {
+            if(mpAutotuneStarted.find(strFuncName) == mpAutotuneStarted.end() || mpAutotuneStarted[strFuncName] == false)
```
evedovelli: This will trigger 2 searches in the map. You could instead call `auto started = mpAutotuneStarted.find(strFuncName)` before the if, and then use the `started` iterator to both test with end() || false.

library/src/include/plancache_autotune.hpp
```diff
@@ -0,0 +1,141 @@
...
+            {
+                mpAutotuneStarted[strFuncName] = true;
+                mpCallCount[strFuncName] = 0;
+                bestSolution[strFuncName] = std::make_pair(1e100, nullptr);
```
evedovelli: Consider replacing 1e100 with `std::numeric_limits<float>::max()`

evedovelli: Same thing when setting the autotune to false below.

library/src/include/plancache_autotune.hpp
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+
+        //Set current solution to try in autotune
+        template <typename T>
+        void setAutotune(const std::string strFuncName, const hiptensorHandle_t handle, const hiptensorPlan_t plan)
```
evedovelli: If you decide to change the maps to use an enum as key, you'd have to change the interface of these functions too.

library/src/plan_cache.cpp
```diff
@@ -0,0 +1,233 @@
...
+    PlanCache::HashId PlanCache::getHashID(hiptensorOperationDescriptor_t desc)
+    {
+        auto tag         = desc->mTag;
+        auto ADataType   = desc->mDescA ? desc->mDescA->mType : hiptensor::NONE_TYPE;
+        auto BDataType   = desc->mDescB ? desc->mDescB->mType : hiptensor::NONE_TYPE;
+        auto CDataType   = desc->mDescC ? desc->mDescC->mType : hiptensor::NONE_TYPE;
+        auto DDataType   = desc->mDescD ? desc->mDescD->mType : hiptensor::NONE_TYPE;
+        auto computeType = desc->mDescCompute;
+        auto mOperationType = desc->mOperationType;
+        auto mContractionOpId = desc->mContractionOpId;
```
evedovelli: Avoid using the prefix `m` for the local variables, otherwise a reader could mistakenly understand them as member attributes.

library/src/plan_cache.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        //Get lengths
+        std::vector<std::size_t> lengthsA = hiptensor::getTensorLengths(desc->mDescA);
+        std::vector<std::size_t> lengthsB = hiptensor::getTensorLengths(desc->mDescB);
+        std::vector<std::size_t> lengthsC = hiptensor::getTensorLengths(desc->mDescC);
+        std::vector<std::size_t> lengthsD = hiptensor::getTensorLengths(desc->mDescD);
+
+        std::vector<std::size_t> arr_lsm;
```
evedovelli: Use camel case for the variable name.

evedovelli: Review the other variables in the change as well. I see a mix of camelCase and snake_case

library/src/plan_cache.cpp
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    PlanCache::Uid PlanCache::getSolutionID(HashId hash_id)
+    {
+        if(mPlanCacheLines.find(hash_id)!=mPlanCacheLines.end()) return mPlanCacheLines[hash_id];
+        return 0ull;
+    }
+
+    hiptensorStatus_t PlanCache::writeFile(const char filename[])
```
evedovelli: The format of the file we write here becomes part of the API and can't be changed from one version to another, unless we break the API (major version updates). Did you consider creating this file using a json format? It'd probably give us more flexibility for small changes and additions.

If you decide to use a json format for the file, there are many good C++ libraries to help on serializing and deserializing a class into a json. I've used https://github.com/nlohmann/json in the past, but there are faster libraries, as https://github.com/Tencent/rapidjson. There licenses are permissive and can be used, but if using the second one, we might need to list it in our license.

library/src/reduction/hiptensor_reduction.cpp
```diff
@@ -336,6 +351,7 @@ hiptensorStatus_t hiptensorReduce(const hiptensorHandle_t handle,
                 int  n     = pSolution->problemDim();
                 auto flops = std::size_t(2) * n;
                 auto bytes = pSolution->problemBytes();
+                std::cout<<flops<<" "<<bytes<<std::endl;
```
evedovelli: Leftover?

test/01_contraction/contraction_test.cpp
```diff
@@ -494,6 +494,18 @@ namespace hiptensor
             CHECK_HIPTENSOR_ERROR(hiptensorCreatePlanPreference(
                 handle, &planPref, algorithm, HIPTENSOR_JIT_MODE_NONE));
 
+            /**************************
+            * Disable Plan Cache for tests
+            ***************************/
+
+            const hiptensorCacheMode_t cacheMode = HIPTENSOR_CACHE_MODE_NONE;
```
evedovelli: I agree on leaving these tests unchanged, but I suggest adding new tests where you enable and validate the plan cache.

evedovelli (commented):
The changes look good, and the functionality appears to work as expected. I left some suggestions on things that I think could improve performance and readability. I strongly recommend adding new tests for the Plan Cache.

library/src/contraction/contraction_selection.cpp
Ryker0627: changed it to std::numeric_limits<float>::max()

library/src/contraction/contraction_selection.cpp
Ryker0627: tmpCandidates[idx] is raw pointer not an object. Is there any benefit to use std::move for raw pointer?

library/src/contraction/hiptensor_contraction.cpp
Ryker0627: changed it to handle->getDevice() and handle->getPlanCache().

library/src/contraction/hiptensor_contraction.cpp
Ryker0627: candidate is a raw pointer not an object. Is there any benefit to use std::move for raw pointer?

library/src/contraction/contraction_selection.cpp
evedovelli: There are no benefits. The time to copy or move a pointer is the same.

library/src/contraction/hiptensor_contraction.cpp
evedovelli: You're right, there are no benefits. Just keep as is.

library/src/contraction/hiptensor_contraction.cpp
Ryker0627: pref->mCandidates is designed to be void* to hold different types of pointer(contraction, permutation etc.). It seems we can not directly process it.

library/src/plan_cache.cpp
Ryker0627: removed frefix m.

library/src/plan_cache.cpp
Ryker0627: Used camel case for it and other variables.

library/src/plan_cache.cpp
Ryker0627: Yes, a json file is good for back compatibility. But the plan cache file is a intermediate temporary file and it is decided by the user whether to write/read it by using our API function hiptensorHandleWritePlanCacheToFile/hiptensorHandleReadPlanCacheFromFile. That  intermediate temporary file should not be used in upgraded version. Also cuTensor is using binary file, so we can keep consistent with it.

library/src/reduction/hiptensor_reduction.cpp
Ryker0627: yes, it is a leftover. removed it. Sorry didn't check it before.

test/01_contraction/contraction_test.cpp
Ryker0627: Added a new test named "plan_cache_test".

library/src/hiptensor.cpp
Ryker0627: renamed it to mKernelRank.

library/src/include/plan_cache.hpp
Ryker0627: Renamed it to UpdatablePriorityQueue.

library/src/include/plan_cache.hpp
Ryker0627: Have renamed all  methods and variables to be  camel case.

library/src/include/plan_cache.hpp
Ryker0627: renamed it to mHeapData

library/src/plan_cache.cpp
evedovelli: Ok, I agree with you.

library/src/include/plan_cache.hpp
Ryker0627: renamed it to mPqUidUsedTimes. We need to frequently update a item in the heap but std::priority_queue does not provide such function. If we use std::priority_queue, we have to use another tempoary heap to store unchanged items and add updated item to the heap. The time complexity is O(nlogn) and is quite slow. So with this UpdatablePriorityQueue, we can directly update the item with time complexity O(logn).

library/src/include/plan_cache.hpp
Ryker0627: already run all related code files with clang-format -i path/file

library/src/include/plan_cache.hpp
evedovelli: Thank you for the explanation. I agree with your solution!

library/src/include/plancache_autotune.hpp
Ryker0627: used a enum class as the key and also renamed all the variables. mpCallCount  count the number of a function call times. mpAutotuneStarted indicate a autotune process already started.

library/src/include/plancache_autotune.hpp
Ryker0627: changed it accordingly.

library/src/include/plancache_autotune.hpp
Ryker0627: changed them accordingly.

library/src/include/plancache_autotune.hpp
Ryker0627: changed them as well.

library/src/contraction/contraction_selection.cpp
```diff
@@ -106,17 +106,17 @@ namespace hiptensor
...
 
         std::string          best_op_name;
-        ContractionSolution* bestSolution = nullptr;
-        PerfMetrics          bestMetrics  = {
-            0,
-            "",
-            0,
-            0,
-            0,
+        ContractionSolution* mBestSolution = nullptr;
```
evedovelli: Here, it'd be best to keep `bestSolution` as it was before, since the variable isn't a member attribute of a class. We can also change this in a next PR.

library/src/reduction/hiptensor_reduction.cpp
evedovelli: No worries! That's why we have code reviews :)

evedovelli (dismissed):
Thanks for all the changes! It looks very good!

library/src/contraction/contraction_selection.cpp
Ryker0627: reverted it. This was an error that occurred when I used VS Code to replace bestSolution with mBestSolution in the PlancacheAutotuneMgr class. Thanks for your keen observation!

evedovelli (dismissed):

evedovelli (approved):

PR #410, opened 2025-10-23
Convert include guards from #ifndef/#define/#endif to #pragma once

evedovelli:
## Motivation

Changing the remaining include files to use `#pragma once`. It makes the compilation faster and avoid naming conflicts and inconsistencies.

## Technical Details

-

## Test Plan

Building the project should not break.

## Test Result

No errors while building the project.

## Submission Checklist


Ryker0627 (approved):
lgtm, thanks for the effort.

PR #411, opened 2025-10-24
remove large tensor sizes causing problem in benchmark

Ryker0627:
## Motivation

1. Removed large tensor sizes from the .yaml files in the bench folder that were causing issues in benchmark tests on MI200 and MI300X.
2. Added cmake-generated file "hiptensor-version.h" to .gitignore (this is not related tensor size problem, just another small fix).

## Test Plan

run all benchmark scripts (*.sh) under "hipTensor/scripts/performance".

## Test Result

All tests has passed.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


test/02_elementwise/configs/bench/rank6_test_params.yaml
```diff
@@ -9,14 +9,14 @@ Alphas:
...
-  - [ 64, 64, 64, 32, 35, 16]
-  - [ 30, 64, 64, 32, 35, 48]
-  - [ 64, 64, 40, 32, 35, 48]
+  - [16, 32, 32, 16, 6, 16]
+  - [23, 36, 36, 18, 6, 20]
+  - [30, 40, 40, 20, 6, 24]
+  - [37, 44, 44, 22, 6, 28]
+  - [44, 48, 48, 24, 6, 32]
+  - [51, 52, 52, 26, 6, 36]
+  - [58, 58, 58, 29, 6, 42]
```
evedovelli, line 18: The downside of changing most of the sets of values from some of the benchmarks is that we won't be able to directly compare with the benchmarks ran for the previous versions.

evedovelli (commented):
LGTM!

evedovelli (approved):

test/02_elementwise/configs/bench/rank6_test_params.yaml
Ryker0627: Yes, I agree. Fortunately, we’re still at the early stage of benchmark testing in hipTensor, since there haven’t been any previous benchmark results. Let’s keep these sizes fixed so we can perform comparison benchmarks in the future.

evedovelli: Sure, I agree with you.

PR #413, opened 2025-10-28
Add option to configure memory layout for tests and benchmarks

evedovelli:
## Motivation

This change adds a Memory Layout field to the configuration YAML used for tests and benchmarks in order to easily allow switching the memory layout in between Row major or Column major in tests and benchmarks to evaluate the performance with both.

## Technical Details

The new `Memory Layout` option can take three values: `HIPTENSOR_MEMORY_LAYOUT_DEFAULT`, `HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR`, and `HIPTENSOR_MEMORY_LAYOUT_ROW_MAJOR`. `HIPTENSOR_MEMORY_LAYOUT_DEFAULT` will use the default layout as set during the library compilation, and the other options will respectively force to column major and row major.

The generated output CSV file includes the strides generated from the chosen memory layout.

## Test Plan

The `Memory Layout` was added to a few tests to validate the multiple operations can correctly handle the different layouts.

## Test Result

All validation tests were executed and are passing.

## Submission Checklist


Ryker0627:
Hi @evedovelli 

Thanks for adding it! This new field makes it much easier for us to switch between different memory layouts.

Could you keep only one item under "Memory Layout", such as HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR? If there are two items, we’ll need to separate the results into row-major and column-major when comparing with cuTENSOR, which complicates things. We only need one layout’s results for the cuTENSOR comparison.

Also, having two items would double the benchmark runtime. I suggest we continue using HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR, as my tests show that the column-major layout performs better for permutation cases. Maybe reduction is same as well.

evedovelli:
> Could you keep only one item under "Memory Layout", such as HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR? If there are two items, we’ll need to separate the results into row-major and column-major when comparing with cuTENSOR, which complicates things. We only need one layout’s results for the cuTENSOR comparison.

@Ryker0627 my idea was indeed run both column and row and separate the results. Given that the performance is very different depending on the layout, I think it's a good idea to keep track of both.

But I understand your concern of doubling the runtime, so I'll keep only the column major for now, and we can review this as we move.

Ryker0627:
> > Could you keep only one item under "Memory Layout", such as HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR? If there are two items, we’ll need to separate the results into row-major and column-major when comparing with cuTENSOR, which complicates things. We only need one layout’s results for the cuTENSOR comparison.
> 
> @Ryker0627 my idea was indeed run both column and row and separate the results. Given that the performance is very different depending on the layout, I think it's a good idea to keep track of both.
> 
> But I understand your concern of doubling the runtime, so I'll keep only the column major for now, and we can review this as we move.

@evedovelli 
Okay, I agree it’s worthwhile to track both layout types. Could you add a column in the output that indicates whether it’s column-major or row-major? It’s not easy to distinguish the results just from the newly added strides output.

There’s an environment variable, HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR, which can set all runs to use either column-major or row-major layout. How is this handled if both the HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR variable and a Memory Layout input are specified in the YAML file?

evedovelli:
> > > Could you keep only one item under "Memory Layout", such as HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR? If there are two items, we’ll need to separate the results into row-major and column-major when comparing with cuTENSOR, which complicates things. We only need one layout’s results for the cuTENSOR comparison.
> > 
> > 
> > @Ryker0627 my idea was indeed run both column and row and separate the results. Given that the performance is very different depending on the layout, I think it's a good idea to keep track of both.
> > But I understand your concern of doubling the runtime, so I'll keep only the column major for now, and we can review this as we move.
> 
> @evedovelli Okay, I agree it’s worthwhile to track both layout types. Could you add a column in the output that indicates whether it’s column-major or row-major? It’s not easy to distinguish the results just from the newly added strides output.

Yes, I can add a column to indicate the layout. I'll make the change.

> There’s an environment variable, HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR, which can set all runs to use either column-major or row-major layout. How is this handled if both the HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR variable and a Memory Layout input are specified in the YAML file?

The memory layout can have three values:
- `HIPTENSOR_MEMORY_LAYOUT_DEFAULT` will use the default layout as set by the env var, or during compilation.
- `HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR` will always force to column.
- `HIPTENSOR_MEMORY_LAYOUT_ROW_MAJOR` will always force row major.

Ryker0627:
> > > > Could you keep only one item under "Memory Layout", such as HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR? If there are two items, we’ll need to separate the results into row-major and column-major when comparing with cuTENSOR, which complicates things. We only need one layout’s results for the cuTENSOR comparison.
> > > 
> > > 
> > > @Ryker0627 my idea was indeed run both column and row and separate the results. Given that the performance is very different depending on the layout, I think it's a good idea to keep track of both.
> > > But I understand your concern of doubling the runtime, so I'll keep only the column major for now, and we can review this as we move.
> > 
> > 
> > @evedovelli Okay, I agree it’s worthwhile to track both layout types. Could you add a column in the output that indicates whether it’s column-major or row-major? It’s not easy to distinguish the results just from the newly added strides output.
> 
> Yes, I can add a column to indicate the layout. I'll make the change.
> 
> > There’s an environment variable, HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR, which can set all runs to use either column-major or row-major layout. How is this handled if both the HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR variable and a Memory Layout input are specified in the YAML file?
> 
> The memory layout can have three values:
> 
> * `HIPTENSOR_MEMORY_LAYOUT_DEFAULT` will use the default layout as set by the env var, or during compilation.
> * `HIPTENSOR_MEMORY_LAYOUT_COLUMN_MAJOR` will always force to column.
> * `HIPTENSOR_MEMORY_LAYOUT_ROW_MAJOR` will always force row major.

@evedovelli 
That sounds good, thanks for the reply!

evedovelli:
@Ryker0627 I changed the code to print in the generated CSV files a column with the Memory Layout.

test/01_contraction/contraction_test.cpp
```diff
@@ -975,6 +982,109 @@ namespace hiptensor
         }
     }
 
+    void ContractionTest::fillStridesIfNeeded(std::vector<std::vector<std::size_t>>&       strides,
```
Ryker0627: A function named hiptensor::stridesFromLengths exists in util.hpp. It seems it is identical to this one.

test/02_elementwise/elementwise_trinary_op_test.cpp
```diff
@@ -332,6 +337,11 @@ namespace hiptensor
         auto beta         = std::get<5>(param);
         auto gamma        = std::get<6>(param);
         auto operators    = std::get<7>(param);
+        auto memoryLayout = std::get<8>(param);
+
+        std::vector<int64_t> stridesIn  = {};
+        std::vector<int64_t> stridesOut = {};
+        fillStridesIfNeeded(stridesIn, stridesOut, lengths, permutedDims, memoryLayout);
```
Ryker0627: In the hiptensorCreateTensorDescriptor function, the strides are computed using stridesFromLengths according to the specified memory layout. Why is fillStridesIfNeeded added for each operation test? Also, what is the reason for separating stridesIn and stridesOut?

test/01_contraction/contraction_test.cpp
evedovelli: Great catch! I'm going to use the existing function.

test/02_elementwise/elementwise_trinary_op_test.cpp
evedovelli: `hiptensorCreateTensorDescriptor` sets the strides using `stridesFromLengths` to memory layout initialized by the envvar `HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR`. So it doesn't provide me with a fine grain control of the stride in use per test, as I need to set the memory layout given by `memoryLayout` instead of the `HIPTENSOR_DEFAULT_STRIDES_COL_MAJOR` general env var.

The reason for separating the `stridesIn` and `stridesOut` is because `stridesIn` are for input tensors A, B and C, which all have the same mode (e.g. `[m, n, k]` with strides `[1, m, m*n]`), and `stridesOut` is for the output tensor D, which can be permuted, so it has a different mode and respectively a different strides (e.g. `[k, m, n]` with strides `[1, k, k*m]`).

Ryker0627: Thanks for the detailed explanation! Is it possible for all the strides to be obtained through the same function, rather than using a new one each time?

evedovelli: I still need to keep some of the logics from the `fillStridesIfNeeded` functions, so I'll need to keep them. But for the stride calculation itself I'm changing to call the `stridesFromLengths` and so I do not need to replicate that function.

test/01_contraction/contraction_test.cpp
evedovelli: I made these changes

Ryker0627 (commented):
Would you consider moving the fillStridesIfNeeded function to common.hpp and common.cpp to prevent duplicating it in the five test .cpp files?

evedovelli:
> Would you consider moving the fillStridesIfNeeded function to common.hpp and common.cpp to prevent duplicating it in the five test .cpp files?

I moved fillStridesIfNeeded to common and made the required adjustments to solve some linkage issues from the change.

Ryker0627 (dismissed):
LGTM, thanks!

Ryker0627 (approved):
lgtm

